PR #4, opened 2019-09-06
Oleksii Demchenko solution split app to the component

alexDemche:


src/components/Nav/Nav.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import './Nav.css';
+
+export default function Nav() {
```
irenhh: no need for `default` here, it's for exporting all the file

PR #20, opened 2019-09-08
Solution

VladGritsenko:


src/components/Header/Header.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import './Header.css';
+import Navigation from '../Navigation/Navigation';
+
+const Header = () =>
+  (
```
YuriiHoliuk: put open brace on the same line with `=>`

PR #27, opened 2019-09-10
Andrii solution

Rafler:
https://rafler.github.io/react_decompose/


src/components/Header/Header.js
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import './Header.css'
+
+function Header() {
+  return (
```
irenhh: rewrite without return

irenhh: apply this to all the components

Rafler: fixed)

PR #90, opened 2020-02-13
Develop

LeonidShv:
https://leonidshv.github.io/react_decompose/


src/components/Article/Article.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import Paragraph from '../Paragraph/Paragraph';
+import Caption from '../Caption/Caption';
+
+export default function Article() {
+  return (
+    <article className="article">
+      <Caption className="article__title" text="Headline" />
```
irenhh: use className for usual elements, no need to pass it to component firstly

irenhh: the same for `text`

src/components/Paragraph/Paragraph.jsx
```diff
@@ -0,0 +1,60 @@
...
+  consectetur ex ut odio tristique vehicula. Ut ligula tortor,
+  tincidunt quis sodales vitae, ornare a turpis. Proin sit amet finibus
+  enim. Fusce tempus a neque vitae tempor. Aenean rutrum, libero
+  iaculis interdum vulputate, dui eros vehicula nisi, at interdum enim
+  lacus eu diam.`,
+];
+
+export default function Paragraph() {
+  return (
+    paragraphes.map(text => <p className="article__paragraph">{text}</p>)
```
irenhh: use just text, no need to put it in an array

PR #149, opened 2020-06-22
solution

tanyamoore:
https://tanyamoore.github.io/react_decompose

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,79 +1,15 @@
 import React from 'react';
 import './App.css';
+import Article from './components/Article/Article';
+import Welcome from './components/Welcome/Welcome';
+import Header from './components/Header/Header';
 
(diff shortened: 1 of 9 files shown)


src/App.js
```diff
@@ -1,79 +1,15 @@
 import React from 'react';
 import './App.css';
+import Article from './components/Article/Article';
+import Welcome from './components/Welcome/Welcome';
+import Header from './components/Header/Header';
```
danheim, line 5: use named exports (optional)

src/components/Article/Article.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import './Article.css';
+
+function Article() {
```
danheim, line 4: arrow functions (every component)

PR #155, opened 2020-06-22
solution

germandolia:
[DEMO LINK](https://germandolia.github.io/react_decompose/)


src/App.js
```diff
@@ -1,79 +1,16 @@
 import React from 'react';
 import './App.css';
 
+import { Welcome } from './components/welcome/Welcome';
+import { Header } from './components/header/Header';
+import { Article } from './components/article/Article';
```
danheim, line 6: Named exports, I like it!

src/components/article/Article.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import './Article.css';
+
+const Article = () => (
```
danheim, line 4: export const

src/App.js
danheim, line 6:
```suggestion
import { Welcome } from './components/Welcome/Welcome';
import { Header } from './components/Header/Header';
import { Article } from './components/Article/Article';
```
https://mateacademy.slack.com/archives/G011BRK1WBH/p1592988955011600

PR #159, opened 2020-06-23
Decompose

tyooma:
[DEMO LINK](https://tyooma.github.io/react_decompose/)


src/App.js
```diff
@@ -1,79 +1,15 @@
 import React from 'react';
 import './App.css';
+import { Article } from './components/article/Article';
+import { Header } from './components/header/Header';
+import { Section } from './components/section/Section';
```
danheim, line 5:
```suggestion
import { Article } from './components/Article/Article';
import { Header } from './components/Header/Header';
import { Section } from './components/Section/Section';
```

PR #411, opened 2021-05-17
solution

provolovskyi-illia:
[DEMO LINK](https://provolovskyi-ilia.github.io/react_decompose/)


rauzok (approved):
good!

VinnieJ-2k20 (changes requested):
Nice solution, but it still needs some adjustments.

Create separate folder for each component and put your `.js` and `.css` files there.
Also add Navigation component.

aliceblunt8 (approved):

PR #417, opened 2021-05-17
Develop

RuslanaPodshyvalova:
[DEMO LINK](https://RuslanaPodshyvalova.github.io/react_decompose/)


provolovskyi-illia (approved):
Good!)

src/components/Article.scss
```diff
@@ -0,0 +1,30 @@
+article {
```
VinnieJ-2k20, line 1: Let's not style using tags, you already have classNames.

VinnieJ-2k20 (changes requested):
Nice solution. Good job with scss.

Let's make some optimisations.

First, create separate folder for each component and put corresponding `js` and `.css` files there.
And one more comment below.

VinnieJ-2k20 (approved):
Good job 👍

PR #434, opened 2021-05-17
solution

KapToH337:
[DEMO LINK](https://KapToH337.github.io/react_decompose/)


src/App.js
```diff
@@ -1,79 +1,16 @@
 import React from 'react';
 import './App.css';
 
+import Welcome from './components/welcome/Welcome';
```
victor-arutiunov, line 4: I think that you can put all components in one file, but it's up to you)

src/components/welcome/Welcome.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
```
victor-arutiunov, line 1: Nice work!

aliceblunt8 (changes requested):
- change the extensions of all files with jsx inside to `.jsx`
- rename all components folders to start with uppercase first letter (like Header, etc)

aliceblunt8 (approved):

PR #440, opened 2021-05-17
solution

den-bulaev:
Demo: https://den-bulaev.github.io/react_decompose/


illisogori (changes requested):
Не все запушил, посомтри git status.

illisogori (approved):
Good job.

src/components/Nav.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+function Nav() {
```
aliceblunt8, line 3:
```suggestion
function Navigation() {
```

aliceblunt8, line 3: import Navigation styles here?

aliceblunt8 (changes requested):
also please create a separate folder for each component inside components folder

PR #542, opened 2021-08-28
solution

dima-tkachenko:


illiarerun:
I can't see all the files. You need to add them to the commit and push again

varseniuk (approved):
For this task let it be, but in future - create a separate component in a separate folder. 
Component `Navigation` should be in another folder (not in the `Header`)

Снимок экрана 2021-08-30 в 15 03 08

PR #548, opened 2021-08-29
React_decompose

zikozakazabiyaka:
[DEMO LINK](https://zikozakazabiyaka.github.io/react_decompose/)


src/components/Article/Article.css
```diff
@@ -0,0 +1,36 @@
...
+.article__paragraph:not(:first-child) {
+  margin-top: 0.7em;
+}
+
+@media (min-width: 500px) {
+  .header {
+    text-align: left;
+    display: flex;
+    justify-content: space-between;
+  }
```
evgen1sider, line 24: this don't needed

evgen1sider (approved):
almost done

src/components/Article/Article.css
varseniuk: Yep. Evgenii is right

varseniuk (changes requested):
Everything is almost good, but let's divide your styles properly

varseniuk (approved):
🔥

PR #553, opened 2021-08-30
React_decompose solved

Andrii-Pinchuk:


Crystallord (approved):
Cool!

varseniuk (changes requested):
Each component and it's styles file should be in a separate directory.
E.g:
`/src/components/Welcome/` directory should contain only two files:
Welcome.jsx
Welcome.scss

Move all your components to separate folders. Now they are all in the `Welcome` folder
Снимок экрана 2021-08-30 в 11 48 42

Andrii-Pinchuk:
> Each component and it's styles file should be in a separate directory.
> E.g:
> `/src/components/Welcome/` directory should contain only two files:
> Welcome.jsx
> Welcome.scss
> 
> Move all your components to separate folders. Now they are all in the `Welcome` folder
> Снимок экрана 2021-08-30 в 11 48 42
Thanks, also I think it's better to put all blocks inside tag 'main' in folder 'main' and then in separate folders, because in app.js also can be 'foooter'

src/App.js
```diff
@@ -1,81 +1,15 @@
 import React from 'react';
+
+import Main from './components/Main/Main';
+
+import './components/Main/Welcome/Welcome.css';
+import './components/Main/Header/Header.css';
+import './components/Main/Article/Article.css';
```
varseniuk, line 7:
```suggestion
```
You should import the styles directly in the component, that uses them.
E.g. you should import `Header.css` in your `Header` component. Not here

varseniuk (changes requested):
One small fix left

varseniuk (approved):

PR #571, opened 2021-08-30
develop

RomSmile:
DEMO LINK: https://RomSmile.github.io/react_decompose/


roman-hado:
не видно усіх твоїх файлів з проекту. Спробуй знову закомітити і запушити але не забувай про команду - git add . -

varseniuk (changes requested):
All is good, but pay attention to the structure:
each component should be in the separate folder (not in `Main`)
E.g. `/components/Welcome`, `/components/Main`, `components/Header` etc.
Let's fix it properly

varseniuk (approved):
'Navigation` should be in a separate folder as well. Pay attention for this in future.It's important because, for example, it may be reused not only in the `Header` but in some other components

PR #612, opened 2021-10-07
solution

caelum29:
[DEMO LINK](https://caelum29.github.io/react_decompose/)

### DIFF
--- a/src/components/Header/Header.jsx
+++ b/src/components/Header/Header.jsx
@@ -1,29 +1,11 @@
 import React from 'react';
+import { Navigation } from './Navigation/Navigation';
 
 export function Header() {
   return (
     <header className="header">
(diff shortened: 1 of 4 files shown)


src/components/Article/Article.scss
```diff
@@ -0,0 +1,36 @@
...
+.article__paragraph:not(:first-child) {
+  margin-top: 0.7em;
+}
+
+@media (min-width: 500px) {
+  .header {
+    text-align: left;
+    display: flex;
+    justify-content: space-between;
+  }
```
Anna-Tkachenko, line 24: Move it to the styles of Header component

Anna-Tkachenko (changes requested):
It's better to use here the same way for declaring each component:
export function ComponentName() { }
or
export const ComponentName = () => ()

PR #659, opened 2021-11-02
My first variant react app

misharosa:
[DEMO LINK](https://misharosa.github.io/react_decompose/)


DionStrzh (approved):
Good one with SCSS

Navigation can be separate component, but now it's fine

DionStrzh (changes requested):
Fix navigation BEM block (create separate component)

VinnieJ-2k20 (approved):
Good job 👍 
`Navigation` files should be in its own folder.

PR #684, opened 2021-12-02
completed react decompose

i-stanko:


Veronika-Korona (approved):
Good job!

src/components/Navbar/Navbar.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import './Navbar.css';
+
+const Navbar = () => (
+  <nav className="navigation">
+    <a className="navigation__link" href="#about">About</a>
```
ihor-jpeg, line 6:
```suggestion
    <a className="navigation__link" href="#about">
      About
    </a>
```

ihor-jpeg (approved):
Good job!

PR #686, opened 2021-12-02
Develop of react_decompose

Dima-Ziakhor:


src/components/Header/Header.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import './Header.css';
+
+function Header() {
+  return (
+    <header className="header">
+      <h1 className="header__title">Site Name</h1>
+      <nav className="navigation">
```
Vitaliy-0, line 8: Navigation is BEM block, put it into another component

Vitaliy-0 (changes requested):
Fix comment and add DEMO link

ihor-jpeg (approved):
Good job!

PR #687, opened 2021-12-02
init

junglq:


src/components/Header/Header.jsx
```diff
@@ -0,0 +1,15 @@
...
+import './Header.css';
+
+const Header = () => (
+  <header className="header">
+    <h1 className="header__title">Site Name</h1>
+    <nav className="navigation">
+      <a className="navigation__link" href="#about">About</a>
+      <a className="navigation__link" href="#services">Services</a>
+      <a className="navigation__link" href="#contact">Contact</a>
+    </nav>
```
msaikun, line 11: You could create a sub folder in components for navigation

msaikun (approved):
Great job! But where is your demo?

src/components/Header/Header.jsx
ihor-jpeg: agreed

ihor-jpeg (approved):
Well done! Nav could indeed be a component on itself

PR #688, opened 2021-12-02
Sticky Header

skachkov23:
[DEMO](https://skachkov23.github.io/react_decompose/)


src/App.js
```diff
@@ -1,80 +1,15 @@
 import React from 'react';
 import './App.css';
-// Move each BEM block to a separate component (file) and import them here
+import Welcome from './components/Welcome/Welcome';
+import Article from './components/Article/Article';
+import Header from './components/Header/Header';
```
Vitaliy-0, line 5: in future you can add in Header folder file index.jsx and in this file import component Header and export him, it can helps do not repeat yourself in path names

src/components/Header/Header.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import './Header.css';
+
+import Navigation from '../Navigation/Navigation';
```
Vitaliy-0, line 4: also repeat in path

Vitaliy-0 (approved):
Good joB!

src/App.js
skachkov23: > 

Thank u, I'll remember.

ihor-jpeg (approved):
Well done!

PR #696, opened 2021-12-03
React page

MagicBox15:


Vitaliy-0 (approved):
Good work

src/components/Articles/Article.jsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import './Article.css';
+
+function Article() {
+  return (
```
aionova, line 5: you can make it an arrow function, if you don't write anything before return

aionova (approved):

PR #697, opened 2021-12-03
Decomposition

Vitalii-Lytvyn01:
[DEMO LINK](https://Vitalii-Lytvyn01.github.io/react_decompose/)


SerhiiKirik (approved):
Well done 🚀

src/Welcome/Welcome.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import './welcome.css';
+
+function Welcome() {
+  return (
```
aionova, line 5: if you don't write anything before return, you can use an arrow function

src/Welcome/welcome.css
```diff
@@ -0,0 +1,14 @@
+.welcome {
```
aionova, line 1: file with styles is usually named the same way as a component "Welcome.css"

aionova (approved):

PR #700, opened 2021-12-03
solution

Nadiyahr:


dlugash (approved):
GJ!

src/components/Article/Article.jsx
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import './Article.css';
+
+function Article() {
+  return (
```
aionova, line 5: if you don't write anything before return, you can use arrow function

aionova (approved):
Nice!

PR #701, opened 2021-12-03
Rect decompose

TatianaMorozova04:
[DEMO LINK](https://TatianaMorozova04.github.io/react_decompose/)


IrynaHerasymchuk (approved):
All good! Just fix one thing: change the extension of the files where you're using JSX from .js to .jsx :)

src/components/Header/Header.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import './Header.css';
+import Navigation from '../Navigation/Navigation';
+
+function Header() {
+  return (
```
aionova, line 6: you can use arrow function if you don't write anything before return

aionova (approved):

PR #743, opened 2022-01-13
Solution

YevheniiKorzun:
[DEMO LINK](https://YevheniiKorzun.github.io/react_decompose/)


src/App.css
```diff
@@ -13,90 +13,3 @@ body {
   margin: 0;
   background: #eee;
```
Rozdorozhnii, line 14: better to use jsx and scss in all cases exept index.js

src/App.js
```diff
@@ -1,80 +1,17 @@
 import React from 'react';
 import './App.css';
```
Rozdorozhnii, line 2: better to use jsx and scss in all cases exept index.js

Rozdorozhnii (changes requested):

Rozdorozhnii (approved):

danheim (approved):

PR #759, opened 2022-01-13
solution

Viktor9r:
https://Viktor9r.github.io/react_decompose/


src/App.css
```diff
@@ -13,90 +13,3 @@ body {
   margin: 0;
```
gnosis-frog, line 13: Can't see components folder ＞﹏＜

gnosis-frog (changes requested):

Olexandr6 (approved):
Good solution 🔥

PR #784, opened 2022-02-10
solution

mykhailoyanush:
[DEMO LINK](https://mykhailoyanush.github.io/react_decompose/)


src/components/Article/Article.css
```diff
@@ -0,0 +1,36 @@
...
+.article__paragraph:not(:first-child) {
+  margin-top: 0.7em;
+}
+
+@media (min-width: 500px) {
+  .header {
+    text-align: left;
+    display: flex;
+    justify-content: space-between;
+  }
```
Anna-Tkachenko, line 24: This styles should be in Header component css file

src/components/Header/Nav/Nav.jsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import './Nav.css';
+
+function Nav() {
```
Anna-Tkachenko, line 4: Use full component names
```suggestion
function Navigation() {
```

Anna-Tkachenko (changes requested):

src/components/Header/Nav/Nav.jsx
mykhailoyanush: but tag named `nav`?

Anna-Tkachenko (approved):
Good job 👍

PR #787, opened 2022-02-10
Develop

pogrebnyakp:
[DEMO LINK](https://pogrebnyakp.github.io/react_decompose/)


src/components/Nav/Nav.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import './Nav.css';
+
+function Nav() {
```
Anna-Tkachenko, line 4: Use full name for components
```suggestion
function Navigation() {
```

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job 👍

PR #823, opened 2022-03-10
separated files

mariiishka:
[DEMO LINK](https://mariiishka.github.io/react_decompose/)


Ihor-Karpyn (changes requested):
Add Header.tsx and Article.tsx

Ihor-Karpyn (approved):

PR #843, opened 2022-03-30
Finish hw.

dant0n:
[DEMO LINK](https://dant0n.github.io/react_decompose/)


src/components/Article/Article.jsx
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import './Article.css';
+
+function Article() {
```
denys-danyliuk, line 4: better to use arrow functions instead

denys-danyliuk (approved):

PR #851, opened 2022-04-21
develop

MariaBondar:
https://mariabondar.github.io/react_decompose/

### DIFF
--- a/src/components/header/header.jsx
+++ b/src/components/header/header.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import Nav from '../nav/nav'
+import Nav from '../nav/nav';
 import './header.css';
 
 function Header() {
--- a/src/components/nav/nav.css
+++ b/src/components/nav/nav.css
@@ -1,0 +1,14 @@
+.navigation__link {
+  display: inline-block;
+  outline: none;
+  text-decoration: none;
+  opacity: 0.7;
+  padding: 0 0.5em;
+  color: black;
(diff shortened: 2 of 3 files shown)


src/App.css
```diff
@@ -14,74 +14,6 @@ body {
...
-  margin: 0;
-  color: #333;
-}
-
-.article__paragraph:not(:first-child) {
-  margin-top: 0.7em;
-}
-
 @media (min-width: 500px) {
   .header {
```
VETALEK, line 18: This must be in `Header.css`

VETALEK: All `@media` queries have to bee in according `.css` block files

src/components/header/header.jsx
```diff
@@ -0,0 +1,17 @@
...
+
+function Header() {
+  return (
+    <header className="header">
+      <h1 className="header__title">Site Name</h1>
+      <nav className="navigation">
+        <a className="navigation__link" href="#about">About</a>
+        <a className="navigation__link" href="#services">Services</a>
+        <a className="navigation__link" href="#contact">Contact</a>
+      </nav>
```
VETALEK, line 12: Looks like `<nav>` is a separate block, it is better to move it in a separate file, as well as it's styles

VETALEK (changes requested):
Works fine, but need some improvements

AlinaMatuschak (approved):
Will be nice to give similar names for folders on components. You have the Welcome folder, but the article, header, and nav you called with lowercase

PR #895, opened 2022-05-01
Develop

Nikitiliok:
https://Nikitiliok.github.io/react_decompose/


src/components/Header/Header.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import './Header.css';
+
+function Header() {
+  return (
+    <header className="header">
+      <h1 className="header__title">Site Name</h1>
+      <nav className="navigation">
```
Yurii837, line 8: Move tags with "navigation" class to another component(and it's css)

Yurii837 (changes requested):

Yurii837 (changes requested):
Good, byt let's make some changes

danylo-onishchuk (approved):

PR #900, opened 2022-05-10
Develop

Krimat:
[DEMO LINK](https://Krimat.github.io/react_decompose/)


src/components/Article/Article.css
```diff
@@ -0,0 +1,29 @@
...
+  margin: 0;
+  color: #333;
+}
+
+.article__paragraph:not(:first-child) {
+  margin-top: 0.7em;
+}
+
+@media (min-width: 500px) {
+  .header {
```
denyslipin, line 20: I think you should remove this block to another file

src/components/Article/Article.css
```diff
@@ -0,0 +1,29 @@
...
+    text-align: left;
```
denyslipin, line 21: I think you should remove this block to another file

denyslipin (changes requested):

PR #910, opened 2022-05-23
solution

vlad-demchuk:
[DEMO LINK](http://vlad-demchuk.github.io/react_decompose/)


src/components/Article/Article.css
```diff
@@ -1 +1,36 @@
...
+  margin: 0;
+  color: #333;
+}
+
+.article__paragraph:not(:first-child) {
+  margin-top: 0.7em;
+}
+
+@media (min-width: 500px) {
+  .header {
```
alexandra-protyanova, line 20: it belongs to header

src/components/Header/Header.css
```diff
@@ -1 +1,34 @@
...
+
+.header__title {
+  font-weight: 600;
+  display: inline;
+  margin: 0;
+  padding: 0;
+  font-size: inherit;
+}
+
+.navigation__link {
```
alexandra-protyanova, line 21: you should replace navigation to app.css

alexandra-protyanova (changes requested):
Fix comments)

PR #918, opened 2022-05-24
solution

yelnikov-andrii:
[DEMO LINK](https://yelnikov-andrii.github.io/react_decompose/)

### DIFF
--- a/src/Header.jsx
+++ b/src/Header.jsx
@@ -1,0 +1,14 @@
+import React from 'react';
+
+const Header = () => (
+  <header className="header">
+    <h1 className="header__title">Site Name</h1>
+    <nav className="navigation">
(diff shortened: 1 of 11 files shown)


src/Header.jsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
```
iwanttosmile4u, line 2: You need to import file with styles for this component. Create Header.css file and replace header classes into this css file and then import it in Header.jsx

src/Welcome.jsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
```
iwanttosmile4u, line 2: You need to import Welcome.css file with styles for Welcome component.

Denchepornyuk (approved):

PR #919, opened 2022-05-24
React

anastasiia-nikita:
[DEMO LINK](https://anastasiia-nikita.github.io/react_decompose/)


src/components/Header/Header.css
```diff
@@ -1 +1,42 @@
...
+
+.header__title {
+  font-weight: 600;
+  display: inline;
+  margin: 0;
+  padding: 0;
+  font-size: inherit;
+}
+
+.navigation__link {
```
dellachaise, line 21: .navigation__link - isn't a part of header block. Please remove it.

dellachaise (changes requested):
Good job!

Please, fix one issue.

src/components/Header/Header.jsx
```diff
@@ -1,6 +1,23 @@
...
+  return (
+    <header className="header">
+      <h1 className="header__title">Site Name</h1>
+      <Navigation />
+    </header>
+  );
+}
 
-// Add a default export statement for Header component to use it in the other files
+function Navigation() {
```
Denchepornyuk, line 13: it is better to create separate file for Navigation

Denchepornyuk (approved):

PR #927, opened 2022-05-24
solution

AndriiRut:
[DEMO LINK](https://AndriiRut.github.io/react_decompose/)


src/components/Article/Article.css
```diff
@@ -1 +1,38 @@
...
+.article__paragraph:not(:first-child) {
+  margin-top: 0.7em;
+}
+
+@media (min-width: 500px) {
+  .header {
+    text-align: left;
+    display: flex;
+    justify-content: space-between;
+  }
```
RomekIvantsiv, line 26: This media should be in header block

src/components/Header/Header.css
```diff
@@ -1 +1,35 @@
...
+  opacity: 0.7;
+  padding: 0 0.5em;
+  color: black;
+  transition: opacity 0.2s ease-in-out;
+}
+
+.navigation__link:hover,
+.navigation__link:focus {
+  opacity: 1;
+}
```
RomekIvantsiv, line 35: These styles shouldn't be in header block

RomekIvantsiv (changes requested):
Good job, just small fixes 😄

src/components/Header/Header.css
AndriiRut: It is allready in it.

src/components/Article/Article.css
AndriiRut: oh! Thank you! will fix it)

src/components/Article/Article.jsx
```diff
@@ -1,6 +1,71 @@
 // Don't forget to import the React library
+import React from 'react';
 // Import a css file containig article styles
-
+import './Article.css';
 // Create an Article function returning the HTML of article block
+const Article = () => (
+  <>
```
Denchepornyuk, line 7: do you need `React.Fragment` here ?)

src/components/Header/Header.jsx
```diff
@@ -1,6 +1,20 @@
 // Don't forget to import the React library
+import React from 'react';
 // Import a css file containig header styles
-
+import './Header.css';
 // Create a Header function returning the HTML of header block
-
+const Header = () => (
+  <>
```
Denchepornyuk, line 7: the same, don't add ` React.Fragment` for each component

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #944, opened 2022-05-27
Develop

MabuzaM:
[DEMO LINK](https://MabuzaM.github.io/react_decompose/)


src/MediaQueries.css
```diff
@@ -0,0 +1,18 @@
...
+    margin: 3.5em auto 0;
+    padding: 2em;
+  }
+}
+
+@media (min-width: 800px) {
+  .article {
+    margin: 3.5em auto;
+  }
+}
```
59MrRobot, line 18: I don't think you need to create a separate file for media queries, you can just put them in their respective component `.css` files

59MrRobot (changes requested):
Good job, just one suggestion

src/MediaQueries.css
meforoma: well, essentially it might be a good idea for the bigger projects, where you would create separate files for
mixins, media, typography, constants etc.
So it's legitimate @MabuzaM

meforoma (approved):

src/MediaQueries.css
MabuzaM: Okay thank you.

PR #954, opened 2022-05-29
Solution

huberciak:
[DEMO LINK](https://huberciak.github.io/react_decompose/)


59MrRobot (approved):
👍

ihor-jpeg (changes requested):
you are using arrow functions in one component and function declaration in other, be consistent

ihor-jpeg (approved):

PR #1010, opened 2022-07-01
Only test pre-commit. There is nothing to check now!!

Andriy3323:


Olexandr6 (changes requested):
I can't see all files in your PR, looks like you haven't committed them

src/components/Article/Article.css
```diff
@@ -1 +1,36 @@
...
+.article__paragraph:not(:first-child) {
+  margin-top: 0.7em;
+}
+
+@media (min-width: 500px) {
+  .header {
+    text-align: left;
+    display: flex;
+    justify-content: space-between;
+  }
```
Olexandr6, line 24: Should be moved to `Header.css` file

Olexandr6 (changes requested):

PR #1013, opened 2022-07-01
ready

SpaceHook:
[DEMO LINK](https://SpaceHook.github.io/react_decompose/)

### DIFF
--- a/src/components/Article/Article.css
+++ b/src/components/Article/Article.css
@@ -1,1 +1,36 @@
-/* Put article styles here */
+.article {
+  margin: 5em auto 0;
+  padding: 1em;
+  font-size: 140%;
+  max-width: 800px;
+  background: white;
+  box-shadow: rgba(0, 0, 0, 0.05) 0 3px 15px;
+}
+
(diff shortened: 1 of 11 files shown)


src/components/Article/Article.css
```diff
@@ -1 +1,36 @@
...
+.article__paragraph:not(:first-child) {
+  margin-top: 0.7em;
+}
+
+@media (min-width: 500px) {
+  .header {
+    text-align: left;
+    display: flex;
+    justify-content: space-between;
+  }
```
aedricson, line 24: Maybe, would be better to split this media rule, that the header styles put in Header.css
but i'm not sure)

aedricson (approved):
gj)

src/components/Article/Article.css
Olexandr6: This media related to `Header.css`, so move it there

Olexandr6 (changes requested):

Olexandr6 (approved):

PR #1056, opened 2022-07-27
Reack elements decomposited

LordAsmodey:
[DEMO LINK](https://LordAsmodey.github.io/react_decompose/)


src/components/Article/Article.css
```diff
@@ -1 +1,30 @@
...
+    margin: 3.5em auto;
+  }
+}
+@media (min-width: 500px) {
+
+  .article {
+    margin: 3.5em auto 0;
+    padding: 2em;
+  }
+}
```
Oleksandr-Pustovit, line 30: I think this will fail. 
In the original css file media500 went before media800 and when screen was 800+ the last one overwrited properties from media500. In your code it will not do that.

Oleksandr-Pustovit (changes requested):
Good job!
Only one thing to check.

src/components/Article/Article.css
LordAsmodey: Fixed!

ivan-kadykalo (approved):
🚀

PR #1097, opened 2022-08-03
Solution

oleg-zolkiewskyi:


src/App.css
```diff
@@ -14,74 +14,6 @@ body {
...
-  color: #333;
-}
-
-.article__paragraph:not(:first-child) {
-  margin-top: 0.7em;
-}
-
 @media (min-width: 500px) {
   .header {
     text-align: left;
```
Anna-Tkachenko, line 19: You should also move styles in media queries to corresponding styles (styles for header to Header.css, ...)

Anna-Tkachenko (changes requested):
Please, add demo link to description of your pull request

alexandra-protyanova (approved):

PR #1129, opened 2022-09-01
Solution

anna-kolmakova:
[DEMO LINK](https://anna-kolmakova.github.io/react_decompose/)


src/components/Article/Article.jsx
```diff
@@ -4,3 +4,72 @@
 // Create an Article function returning the HTML of article block
 
 // Add a default export statement for Article component to use it in the other files
+
+import React from 'react';
+import './Article.css';
+
+const Article = () => (
+  <React.Fragment>
```
AndreyJigijalkin, line 12:
```suggestion
  <>
```
You can use empty braces instead of using react.fragment:)

AndreyJigijalkin (approved):
Very good:)

src/components/Article/Article.jsx
AndriiRut: but in this case, you don't need React.Fragment at all!
you already have one parent element and it's `<article></article>`.

src/components/Welcome/Welcome.jsx
```diff
@@ -4,3 +4,16 @@
...
 // Add a default export statement for Welcome component to use it in the other files
+import React from 'react';
+import './Welcome.css';
+
+const Welcome = () => (
+  <React.Fragment>
+    <section className="welcome">
+      <span className="welcome__text">Sticky Header!</span>
+    </section>
+  </React.Fragment>
```
AndriiRut, line 15:
```suggestion
    <section className="welcome">
      <span className="welcome__text">Sticky Header!</span>
    </section>
```

AndriiRut (changes requested):

src/components/Welcome/Welcome.jsx
anna-kolmakova: спасибо, исправила)

AndriiRut (approved):
🚀

PR #1130, opened 2022-09-01
Solution

kir8mir:
- [DEMO LINK](https://kir8mir.github.io/react_decompose/)


src/App.jsx
```diff
@@ -1,81 +1,28 @@
...
-    </main>
+
+      <main className="app">
+        <section className="welcome">
+          <Welcome />
+        </section>
+
+        <article className="article">
+          <Article />
+        </article>
```
OlehMy, line 23:
```suggestion
        <Welcome />
        <Article />
```
In my opinion, better to write these tags directly in the block)

src/components/Navigation/Navigation.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import './Navigation.css';
+
+const Navigation = () => (
+  <nav className="navigation">
+    <a className="navigation__link" href="#about">About</a>
+    <a className="navigation__link" href="#services">Services</a>
+    <a className="navigation__link" href="#contact">Contact</a>
```
OlehMy, line 8:
```suggestion
      {links.map(link => (
        <a
          className="navigation__link"
          key={link.href}
          href={link.href}
        >
          {link.label}
        </a>
      ))}
```
Also in this case you can create some array, and to use map to avoid code duplication))

OlehMy (approved):
Good job)

danylo-onishchuk (approved):

PR #1132, opened 2022-09-02
solution

pelahiia:
[[DEMO LINK](https://pelahiia.github.io/react_decompose/)]


src/App.jsx
```diff
@@ -1,80 +1,16 @@
...
-          turpis. Vestibulum rhoncus placerat nulla vitae suscipit. Curabitur
-          consectetur ex ut odio tristique vehicula. Ut ligula tortor,
-          tincidunt quis sodales vitae, ornare a turpis. Proin sit amet finibus
-          enim. Fusce tempus a neque vitae tempor. Aenean rutrum, libero
-          iaculis interdum vulputate, dui eros vehicula nisi, at interdum enim
-          lacus eu diam.
-        </p>
-      </article>
+      <Welcome />
+      <Header />
```
qoa11a, line 12: It's better to move header outside of \<main\> tag.

qoa11a (changes requested):

src/App.jsx
pelahiia: The problem is that in the given markup <header> is inside <main>
Main has its own classes, so all styles will be changed ii I move it outside the tag
image

src/App.jsx
```diff
@@ -1,80 +1,16 @@
 import React from 'react';
 import './App.css';
+import Welcome from './components/Welcome/Welcome';
+import Header from './components/Header/Header';
+import Article from './components/Article/Article';
 // Move each BEM block to a separate component (file) and import them here
```
liliia-livandovskaya, line 6: remove comments

PR #1149, opened 2022-09-05
Solution

ShmelyovOleg:
[DEMO LINK](https://shmelyovOleg.github.io/react_decompose/)


src/components/Header/Header.jsx
```diff
@@ -1,6 +1,13 @@
-// import the React library here
-// import a css file containig article styles
+import React from 'react';
+import Navigation from '../Navigation/Navigation';
```
oksana-logos-frontend, line 2: Where is your Navigation?It should be in folder components but i didn't find it here  and Git`s check says the same.  
Знімок екрана 2022-09-05 о 22 35 12

oksana-logos-frontend (changes requested):
Almost great)))
But please add Navigation folder with file to components.You import Navigation but I didn't find it.

oksana-logos-frontend (approved):
Great))

danylo-onishchuk (approved):

PR #1154, opened 2022-09-05
solutionReactDecompose

YuriiPronin:


src/components/Welcome/Welcome.jsx
```diff
@@ -1,6 +1,15 @@
...
+import './Welcome.css';
 // Create a Welcome function returning the HTML of welcome block
-
+function Welcome() {
+  return (
+    <section className="welcome">
+      <span className="welcome__text">Sticky Header!</span>
+    </section>
+  );
+}
```
ibondrk, line 12: You can use here function expression as i know it's just more common version and u don't use function body in this case so ... In other files u can do the same)
```
const Welcome = () => (
    <section className="welcome">
    </section>
  ); 
```

ibondrk (approved):

danylo-onishchuk (changes requested):
You may also create a folder for the Navigation component and move navigation there.

danylo-onishchuk (approved):

PR #1170, opened 2022-09-05
Solution

DimaKobzar7:
[DEMO LINK](https://DimaKobzar7.github.io/react_decompose/)


oksana-logos-frontend (approved):
Good👍🏼

src/components/Header/Header.css
```diff
@@ -1 +1,42 @@
...
+
+.header__title {
+  font-weight: 600;
+  display: inline;
+  margin: 0;
+  padding: 0;
+  font-size: inherit;
+}
+
+.navigation__link {
```
AndriiRut, line 21: `<Navigation /> should be the component to)

AndriiRut (changes requested):

src/components/Header/Header.css
DimaKobzar7: navigation should become a separate component along with its own styles?

DimaKobzar7: The assignment says  "You may also create a folder for the `Navigation` component and move navigation there." 
That is, I can make it a component or not.

AndriiRut (approved):
🚀

PR #1208, opened 2022-10-01
Solution

MaxManis:


src/components/Article/Article.jsx
```diff
@@ -1,4 +1,74 @@
-// import the React library here
+import React from 'react';
+import './Article.css';
+
+function Article() {
+  return (
```
OleksiiBlezniuk, line 5: Usually we use arrow functions for components

src/components/Article/Article.jsx
```diff
@@ -1,4 +1,74 @@
...
+        iaculis interdum vulputate, dui eros vehicula nisi, at interdum enim
+        lacus eu diam.
+      </p>
+    </article>
+  );
+}
+
+export default Article;
+
+// export default Article;
```
OleksiiBlezniuk, line 70: seems that this comment should be deleted

src/components/Article/Article.jsx
OleksiiBlezniuk: Also you can export it here

OleksiiBlezniuk, line 5:
```suggestion
export const Article = () => (
```

OleksiiBlezniuk (approved):

PR #1216, opened 2022-10-04
solution

Shepeha-Andriy:
https://Shepeha-Andriy.github.io/react_decompose/


src/components/Header/Header.css
```diff
@@ -1 +1,42 @@
...
+  opacity: 0.7;
+  padding: 0 0.5em;
+  color: black;
+  transition: opacity 0.2s ease-in-out;
+}
+
+.navigation__link:hover,
+.navigation__link:focus {
+  opacity: 1;
+}
```
anton-iskryk, line 34: It's better to put code that is connected with 'navigation' into a separate Navigation component.

src/components/Article/Article.css
```diff
@@ -1 +1,36 @@
...
+.article__paragraph:not(:first-child) {
+  margin-top: 0.7em;
+}
+
+@media (min-width: 500px) {
+  .header {
+    text-align: left;
+    display: flex;
+    justify-content: space-between;
+  }
```
anton-iskryk, line 24: This code should not be here. Move it to the header CSS part, not the article one.

src/components/Header/Header.jsx
```diff
@@ -1,6 +1,15 @@
...
 
-// Create a Header function returning the HTML of header block
+const Header = () => (
+  <header className="header">
+    <h1 className="header__title">Site Name</h1>
+    <nav className="navigation">
+      <a className="navigation__link" href="#about">About</a>
+      <a className="navigation__link" href="#services">Services</a>
+      <a className="navigation__link" href="#contact">Contact</a>
+    </nav>
```
anton-iskryk, line 11: You should replace this code with a link for the separate Navigation component.

anton-iskryk (changes requested):
Good job! But I found some things to change in your code. Keep going!

Denchepornyuk (approved):

PR #1231, opened 2022-10-08
Decompose task completed App.jsx and App.css files were decompose to …

Dikey945:
…separate components
[DEMO LINK](https://%3Cyour_account%3E.github.io/react_decompose/)


src/components/Header/Header.jsx
```diff
@@ -4,3 +4,20 @@
...
+
+function Header() {
+  return (
+    <header className="header">
+      <h1 className="header__title">Site Name</h1>
+      <nav className="navigation">
+        <a className="navigation__link" href="#about">About</a>
+        <a className="navigation__link" href="#services">Services</a>
+        <a className="navigation__link" href="#contact">Contact</a>
+      </nav>
```
myroslav-diiak, line 18: Nav should be separated as component

myroslav-diiak (approved):

src/components/Header/Header.jsx
Dikey945: That is a non-mandatory task

Andr1yK: > That is a non-mandatory task

Yep, but let's try :)

Andr1yK (approved):

PR #1234, opened 2022-10-09
React decompose

Anna-Borodaienko:
[DEMO LINK](https://<Anna-Borodaienko>.github.io/react_decompose/)


src/Navigation/Navigation.jsx
```diff
@@ -0,0 +1,14 @@
...
+
+function Navigation() {
+  return (
+    <nav className="navigation">
+      <a className="navigation__link" href="#about">About</a>
+      <a className="navigation__link" href="#services">Services</a>
+      <a className="navigation__link" href="#contact">Contact</a>
+    </nav>
+  );
+}
```
imaksweb, line 12: I'm not sure, but I think it's better to use arrow function
```suggestion
const Navigation = () => (
  <nav className="navigation">
    <a className="navigation__link" href="#about">About</a>
    <a className="navigation__link" href="#services">Services</a>
    <a className="navigation__link" href="#contact">Contact</a>
  </nav>
);
```

imaksweb (approved):
Good job 👍

src/Navigation/Navigation.jsx
naziamoff: Ilya is right, we usually use arrow functions. What's more, you can add export right here
```suggestion
export const Navigation = () => (
    <nav className="navigation">
      <a className="navigation__link" href="#about">About</a>
      <a className="navigation__link" href="#services">Services</a>
      <a className="navigation__link" href="#contact">Contact</a>
    </nav>
)
```

naziamoff (approved):

PR #1243, opened 2022-10-09
solution

myroslav-diiak:


vladimir-dublya (approved):

src/components/Article/Article.jsx
```diff
@@ -1,6 +1,68 @@
-// import the React library here
-// import a css file containig article styles
+import React from 'react';
+import './Article.css';
 
-// Create an Article function returning the HTML of article block
+function Article() {
```
naziamoff, line 4: better to use arrow function here

naziamoff (approved):
GJ 🔥

PR #1245, opened 2022-10-10
Splited App.js into components

Yeliseiev:


Maksym845 (approved):
Good solution, I didn`t add a Navigation block to my solution, because it was not in the task, but you made it better!

Denchepornyuk (approved):

PR #1250, opened 2022-10-10
Created separate file for each blocks

jojnajbude:


Vladosik1337 (approved):
Good

src/components/Article/Article.jsx
```diff
@@ -1,6 +1,70 @@
-// import the React library here
-// import a css file containig article styles
+import React from 'react';
+
+import './Article.css';
 
 // Create an Article function returning the HTML of article block
+function Article() {
```
naziamoff, line 6: it is better to use arrow functions for functional components

naziamoff (approved):
🔥

PR #1253, opened 2022-10-10
Solved. Ver.1.0

PaulVoron:
[DEMO LINK](https://PaulVoron.github.io/react_decompose/)


src/components/Header/Header.css
```diff
@@ -1 +1,27 @@
...
+  font-size: inherit;
+}
+
+@media (min-width: 500px) {
+  .header {
+    text-align: left;
+    display: flex;
+    justify-content: space-between;
+  }
+}
```
polosanya, line 27: It would be better to place your CSS properties in the right order.
Because for now all of them are mixed and it is harder to divide typography properties from position etc.
Please check all your CSS files and fix them
https://9elements.com/css-rule-order/

src/components/Header/Header.jsx
```diff
@@ -1,6 +1,14 @@
-// import the React library here
 // import a css file containig article styles
 
-// Create a Header function returning the HTML of header block
+import React from 'react';
+import Navigation from '../Navigation/Navigation';
```
polosanya, line 4:
```suggestion
import Navigation from '../Navigation/Navigation.tsx';
```
Please check for errors before commiting. In most cases, they directly say what the problem is

polosanya (changes requested):
Nice start! But unfortunately, it is a critical error with Navigation component now. Also, I don't see it on GitHub, probably you didn't add it to your commit. Please fix it and check my comments. They might be helpful to make your code look cleaner)

polosanya (approved):
Good job!

PR #1261, opened 2022-10-10
Solution

Oleksandra-Shevchenko:


src/components/Article/Article.css
```diff
@@ -1 +1,38 @@
...
+  margin: 0;
+  color: #333;
+}
+
+.article__paragraph:not(:first-child) {
+  margin-top: 0.7em;
+}
+
+@media (min-width: 500px) {
+  .header {
```
Yevheniya-Deryka, line 22: You should better move media for header to header.css

Yevheniya-Deryka (approved):
GJ! 
Just delete header media from article)

Denchepornyuk (approved):

PR #1277, opened 2022-10-13
Solution

AniaHna:
[DEMO LINK](https://AniaHna.github.io/react_decompose/)


src/App.css
```diff
@@ -17,90 +17,3 @@ body {
...
-.article__paragraph:not(:first-child) {
-  margin-top: 0.7em;
-}
-
-@media (min-width: 500px) {
-  .header {
-    text-align: left;
-    display: flex;
-    justify-content: space-between;
-  }
```
Anna-Borodaienko, line 94: `@media (min-width: 500px) {
  .header {
    text-align: left;
    display: flex;
    justify-content: space-between;
  }`

You lost this part of Header styles

Anna-Borodaienko (changes requested):

Anna-Borodaienko (approved):

src/components/Header/Header.jsx
```diff
@@ -4,3 +4,17 @@
 // Create a Header function returning the HTML of header block
 
 // Add a default export statement for Header component to use it in the other files
+import React from 'react';
+import './Header.css';
+import Navigation from '../Navigation/Navigation';
+
+function Header() {
```
naziamoff, line 11: let's use arrow functions for the functional components

naziamoff: + you can return right away in that case

AniaHna: +

naziamoff (approved):
🔥

src/components/Header/Header.jsx
naziamoff: I was expecting it to be a `+` sign actually 😄

AniaHna: yes it was plus))

PR #1299, opened 2022-10-26
react decompose done v1

JeniaGaidukov:
[DEMO LINK](JeniaGaidukov.github.io/react_decompose/)


vlad-demchuk (changes requested):
while scrolling. fix it
image

TarasHoliuk (approved):

PR #1348, opened 2022-11-14
develop

taraselnik:


src/components/Article/Article.css
```diff
@@ -1 +1,30 @@
...
+    margin: 3.5em auto 0;
+    padding: 2em;
+  }
+}
+
+@media (min-width: 800px) {
+  .article {
+    margin: 3.5em auto;
+  }
+}
```
Stepico, line 30: It's better to write media queries right after the class they are related to

src/App.css
```diff
@@ -103,4 +103,4 @@ body {
   .article {
     margin: 3.5em auto;
   }
-}
```
Stepico, line 106: I guess it should be deleted

taraselnik: ок

src/components/Article/Article.css
taraselnik: ok

illiarerun (approved):
GJ

PR #1350, opened 2022-11-14
complete

spiritkyiv:
[DEMO LINK](https://spiritkyiv.github.io/react_decompose/)


taraselnik (commented):
make navigation

src/components/Article/Article.css
```diff
@@ -1 +1,30 @@
...
+    margin: 3.5em auto 0;
+    padding: 2em;
+  }
+}
+
+@media (min-width: 800px) {
+  .article {
+    margin: 3.5em auto;
+  }
+}
```
Anastasiia-Smirnova, line 30: You should move these @media right after .article

src/components/Header/Header.css
```diff
@@ -1 +1,27 @@
...
+  padding: 0;
+  font-size: inherit;
+}
+
+@media (min-width: 500px) {
+  .header {
+    text-align: left;
+    display: flex;
+    justify-content: space-between;
+  }
```
Anastasiia-Smirnova, line 26: Same here, this @media is for .header, so it should be put right after it

src/components/Header/Header.jsx
```diff
@@ -1,6 +1,12 @@
-// import the React library here
-// import a css file containig article styles
+import React from 'react';
+import Navigation from './Navigation/Navigation';
```
Anastasiia-Smirnova, line 2: As for these failures: 'Unable to resolve path to module './Navigation/Navigation'' and 'Missing file extension for "./Navigation/Navigation" '
You should run 'git add .' command and then commit and push

Anastasiia-Smirnova (changes requested):
Nice job! But you should add a few little changes)

spiritkyiv:
> make navigation

You may replace?

spiritkyiv:
I don`t know why test on github failure, on my pc npm lint and npm test shows passed tests

src/components/Header/Header.jsx
```diff
@@ -1,6 +1,12 @@
-// import the React library here
-// import a css file containig article styles
+import React from 'react';
+import Navigation from '../Navigation/Navigation';
```
denys-danyliuk, line 2: probably you forgot to commit file with navigation that's why you have this errors

denys-danyliuk (changes requested):
commit files with navigation

Olexandr6 (approved):

PR #1357, opened 2022-11-15
Task completed

vadymbaranov:


src/components/Article/Article.css
```diff
@@ -1 +1,30 @@
...
+    margin: 3.5em auto 0;
+    padding: 2em;
+  }
+}
+
+@media (min-width: 800px) {
+  .article {
+    margin: 3.5em auto;
+  }
+}
```
kryvets-vadym, line 30: it would be better to move these tags under the artiсle class

src/components/Header/Header.css
```diff
@@ -1 +1,27 @@
...
+  font-size: inherit;
+}
+
+@media (min-width: 500px) {
+  .header {
+    text-align: left;
+    display: flex;
+    justify-content: space-between;
+  }
+}
```
kryvets-vadym, line 27: same here

kryvets-vadym (approved):
GJ!

src/components/Article/Article.css
mykola-koval, line 30: add media rules right after the classes you're applying them to

mkyloa (approved):

PR #1359, opened 2022-11-15
add solution

AlexTsimba:


src/components/Article/Article.css
```diff
@@ -1 +1,30 @@
...
+    margin: 3.5em auto 0;
+    padding: 2em;
+  }
+}
+
+@media (min-width: 800px) {
+  .article {
+    margin: 3.5em auto;
+  }
+}
```
Largoxoxo, line 30: You should add media rule right after the class you are applying it to.

src/components/Header/Header.css
```diff
@@ -1 +1,28 @@
...
+  font-size: inherit;
+}
+
+@media (min-width: 500px) {
+  .header {
+    text-align: left;
+    display: flex;
+    justify-content: space-between;
+  }
+}
```
Largoxoxo, line 28: Same here.

Largoxoxo (approved):
GJ!

src/components/Article/Article.css
illiarerun, line 30: put media rules right after the class that you apply them to

src/components/Header/Header.css
illiarerun: same here

illiarerun (changes requested):
Please add the DEMO LINK to your PR's description and fix these:

PR #1389, opened 2022-12-07
Develop

JOWISSA:


TarasHoliuk:
That is not what we expected (fix positioning):
image

src/components/Header/Header.jsx
```diff
@@ -4,3 +4,18 @@
 // Create a Header function returning the HTML of header block
 
 // Add a default export statement for Header component to use it in the other files
+
+import React from 'react';
+import Navigation from '../Navigation/Navigation';
```
tetiana-ivanchenko, line 9: you still didn't add Navigation component

src/components/Welcome/Welcome.jsx
```diff
@@ -4,3 +4,15 @@
 // Create a Welcome function returning the HTML of welcome block
 
 // Add a default export statement for Welcome component to use it in the other files
```
tetiana-ivanchenko, line 6: remove comments from your code

tetiana-ivanchenko (changes requested):

src/components/Article/Article.css
```diff
@@ -1 +1,36 @@
...
+  margin: 0;
+  color: #333;
+}
+
+.article__paragraph:not(:first-child) {
+  margin-top: 0.7em;
+}
+
+@media (min-width: 500px) {
+  .header {
```
AlisaVlasova, line 20: move header into header component

src/components/Header/Header.jsx
AlisaVlasova: still not here(
if there is any problem fell free to ask in slack

AlisaVlasova (changes requested):

src/components/Welcome/Welcome.css
```diff
@@ -1 +1,15 @@
-/* Put welcome styles here */
+
+.welcome {
```
tetiana-ivanchenko, line 2:
```suggestion
.welcome {
```

tetiana-ivanchenko (approved):

PR #1402, opened 2022-12-18
decomposed app

nikachu404:


anastasiia-roiek (changes requested):
it would be nice to move Navigation into a separate folder

src/components/Article/Article.jsx
```diff
@@ -1,6 +1,68 @@
-// import the React library here
-// import a css file containig article styles
+import React from 'react';
+import './Article.css';
 
-// Create an Article function returning the HTML of article block
+const Article = () => (
+  <>
```
solaryasha, line 5: We don't need this fragment

src/components/Navigation/Navigation.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import './Navigation.css';
+
+export const Navigation = () => (
+  <>
```
solaryasha, line 5: We don't need this fragment

solaryasha (changes requested):

src/components/Article/Article.jsx
solaryasha, line 5: This fragment is useless, cause your component has one root tag - `article`

solaryasha (approved):

PR #1421, opened 2022-12-21
Solution

nurdanokcu:
[DEMO LINK](https://nurdanokcu.github.io/react_decompose/)


UmizDemud (approved):
I'm suprised by the 
```js
function Component {
  return
}
```
syntax, I'm used to
```js
const Component => (
  ...
)
```
or
```js
const Component => {
  return ();
}
```
Not that it matters 😄 , well done 🚀

src/App.jsx
```diff
@@ -1,80 +1,16 @@
 import React from 'react';
 import './App.css';
+import Welcome from './components/Welcome/Welcome';
+import Header from './components/Header/Header';
+import Article from './components/Article/Article';
```
meforoma, line 5: creating `index.` files would shorten the import statements
```suggestion
import Welcome from './components/Welcome';
import Header from './components/Header';
import Article from './components/Article';
```

meforoma (approved):
✅

PR #1463, opened 2022-12-25
splitted files into blocks

alaaashkar:
https://alaaashkar.github.io/react_decompose/


src/App.jsx
```diff
@@ -1,82 +1,17 @@
...
-        </p>
-      </article>
-    </main>
-  );
+  <main className="app">
+    <Welcome />
+    <Header />
+    <Article />
+    <Navigation />
+  </main>;
```
IoghiStudio, line 14: You have to add this in the return() paranthesis and also navigation block inside header block

src/components/Article/Article.jsx
```diff
@@ -1,6 +1,66 @@
-// import the React library here
-// import a css file containig article styles
+import React from 'react';
+import './Article.css';
 
-// Create an Article function returning the HTML of article block
+const Article = () => {
+  <article className="article">
```
IoghiStudio, line 5: use '(' instead of '{'

src/components/Welcome/Welcome.jsx
```diff
@@ -1,6 +1,12 @@
...
+import React from 'react';
+import './Welcome.css';
 
-// Create a Welcome function returning the HTML of welcome block
+function Welcome() {
+  <main className="app">
+    <section className="welcome">
+      <span className="welcome__text">Sticky Header!</span>
+    </section>
+  </main>;
```
IoghiStudio, line 9: you have to return it like 

```
return(
<main className="app">
    <section className="welcome">
    </section>
  </main>;
)
```
)

IoghiStudio (changes requested):

IoghiStudio (approved):
Awesome!

naziamoff (changes requested):
let's make our components arrow functions

src/App.jsx
```diff
@@ -1,82 +1,15 @@
...
-          tincidunt quis sodales vitae, ornare a turpis. Proin sit amet finibus
-          enim. Fusce tempus a neque vitae tempor. Aenean rutrum, libero
-          iaculis interdum vulputate, dui eros vehicula nisi, at interdum enim
-          lacus eu diam.
-        </p>
-      </article>
-    </main>
-  );
-}
+const App = () => {
```
naziamoff, line 7: in the arrow functions you can return right away if you use round braces instead of curly ones. That is better if you don't calculate anything, but need to render sth right away

naziamoff (approved):

PR #1470, opened 2022-12-26
solution

lukacsja:
[DEMO LINK](https://lukacsja.github.io/react_decompose/)


IoghiStudio (approved):
Perfect!

src/components/Article/Article.jsx
```diff
@@ -1,6 +1,68 @@
-// import the React library here
-// import a css file containig article styles
+import React from 'react';
+import './Article.css';
 
-// Create an Article function returning the HTML of article block
+function Article() {
+  return (
```
naziamoff, line 5: would be better to make those an arrow ones

naziamoff (approved):

PR #1474, opened 2022-12-27
Solution

Iksich:
[DEMO LINK](https://Iksich.github.io/react_decompose/)


src/components/Navigation/Navigation.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import './Navigation.css';
+
+function Nav() {
```
naziamoff, line 4: why not an arrow one?

naziamoff (approved):
LGTM
But it is better to add export to your component and make it arrow function.

PR #1477, opened 2022-12-28
react solution

katya-suvorova:


src/components/Article/Article.css
```diff
@@ -1 +1,33 @@
...
+      padding: 2em;
+    }
+  }
+  
+  @media (min-width: 800px) {
+    .article {
+      margin: 3.5em auto;
+    }
+  }
+
```
loralevitska, line 33: add [empty line](https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file) in the end of file, fix it everywhere

loralevitska (changes requested):
Please deploy your solution, and before review check demo_link

src/components/Article/Article.css
katya-suvorova: updated

katya-suvorova: https://katya-suvorova.github.io/react_decompose/

loralevitska (approved):

PR #1484, opened 2023-01-06
final

KareemAchkar:
(https://KareemAchkar.github.io/react_decompose/)


src/components/Article/Article.jsx
```diff
@@ -1,6 +1,68 @@
-// import the React library here
-// import a css file containig article styles
+import React from 'react';
+import './Article.css';
 
-// Create an Article function returning the HTML of article block
+function Article() {
```
naziamoff, line 4: let's make functional components arrow functions

naziamoff (changes requested):
please check whether your task is deployed. Usually I reject without reviewing if I see the demo's not working

src/components/Article/Article.jsx
KareemAchkar: not passing when i am using arrow , doesnt pass it on npm test 
Untitled

PR #1486, opened 2023-01-08
solution

illia-kyselov:
[DEMO LINK](https://illia-kyselov.github.io/react_decompose/)


src/components/Header/Header.jsx
```diff
@@ -1,6 +1,12 @@
-// import the React library here
-// import a css file containig article styles
+import React from 'react';
+import Navigation from '../Navigation/Navigation';
```
tetiana-ivanchenko, line 2: you didn't add Navigation

tetiana-ivanchenko (changes requested):

src/components/Header/Header.jsx
tetiana-ivanchenko: there is your changes, it doesn't include Navigation file
image

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #1497, opened 2023-01-15
react_decompose

dima-korovai:
[DEMO LINK](https://dima-korovai.github.io/react_decompose/)


polosanya (changes requested):
Please add a demo link to your PR description so that we can check your work properly)
Also, let's fix this problem to pass the tests
image

src/components/Article/Article.css
```diff
@@ -1 +1,31 @@
...
+    padding: 2em;
+  }
+}
+
+@media (min-width: 800px) {
+  .article {
+    margin: 3.5em auto;
+  }
+}
+
```
polosanya, line 31:
```suggestion
}
```
The problem is still present.
Leave only one newline in the end
image

polosanya (changes requested):
Lint problem is not fixed. 
Feel free to ask for help in the chat if you have any difficulties

src/components/Article/Article.css
dima-korovai: Hope now it's ok. At least run lint doesn't show up any error.

loralevitska (approved):
Good job!

PR #1499, opened 2023-01-16
NVyrsta

NVyrsta:
[DEMO LINK](https://NVyrsta.github.io/react_decompose/)


OlhaYevstifieieva (approved):
GJ!

src/components/Article/Article.jsx
```diff
@@ -1,6 +1,17 @@
...
+import { articles } from './articles';
 
-// Add a default export statement for Article component to use it in the other files
+const Article = () => (
+  <article className="article">
+    <h1 className="article__title">Headline</h1>
+    {articles.map(([ind, article]) => (
+      <p className="article__paragraph" key={ind}>
+        {article}
+      </p>
```
ArturErmolenko, line 12: when you use `map` it is better use` ul li`, in this task you need make component `Article` and just use it, don't make things difficult :)

ArturErmolenko (changes requested):

ArturErmolenko (approved):
Nice 👍

PR #1535, opened 2023-01-22
React_Decompose were implemented

Kanzatuwka:
[DEMO LINK](https://kanzatuwka.github.io/react_decompose/)


src/components/Header/Header.jsx
```diff
@@ -1,6 +1,16 @@
 // import the React library here
```
tetiana-ivanchenko, line 1: remove all comments

tetiana-ivanchenko (changes requested):

TarasHoliuk (approved):

PR #1550, opened 2023-01-22
Added solution

burovetskyi:
[DEMO LINK](https://burovetskyi.github.io/react_decompose/)
https://prnt.sc/SZW77FD1a-7m
Do not add new commit , but all tests are allright.


NVyrsta (changes requested):
You specify the path to the component Navigation, which you did not add to the project.
Maybe you immediately make a git commit and forget to make a git add before that?
Try to solve the problem with the tests

PR #1553, opened 2023-01-23
Split components

Coyoote:
[DEMO LINK](https://Coyoote.github.io/react_decompose/)


Anton-Zhytonbaiev (approved):
Nice code!)

src/App.jsx
```diff
@@ -1,81 +1,18 @@
...
-        </p>
-      </article>
-    </main>
+    <>
+      <Welcome />
+      <Header />
+      <main className="app">
+        <Article />
+      </main>
+    </>
```
alexandra-protyanova, line 15: What is the point to make the structure so complicated? Just wrap your 3 components with `main` tag, it'll work in the same way)

alexandra-protyanova (approved):
GJ 🌟

src/App.jsx
Coyoote: Yesterday Denys was breaking down this task and said that the Header and Welcome components should be above the `main` tag 😅 and that's why I did it like that.

alexandra-protyanova: In my opinion, we are just breaking previous code to components, but okay you can leave it as it is)

PR #1564, opened 2023-01-24
sol 1.0

aliliolek:
https://aliliolek.github.io/react_decompose/


src/components/Navigation/Navigation.jsx
```diff
@@ -0,0 +1,14 @@
+// import the React library here
+import React from 'react';
+// import a css file containig article styles
```
polosanya, line 3: You can delete the comments

polosanya (approved):
Good job!

PR #1624, opened 2023-03-01
result

novak-k:
[DEMO LINK](https://nobak-k.github.io/react_decompose/)


src/components/Navigation/Navigation.jsx
```diff
@@ -0,0 +1,20 @@
+// import the React library here
+// import a css file containig article styles
+
+// Create a Header function returning the HTML of header block
+
+// Add a default export statement for Header component to use it in the other files
```
nazarmatsevych, line 6: Remove your and default comments

nazarmatsevych (changes requested):

src/components/Navigation/Navigation.jsx
novak-k: done

IvanFesenko (approved):
Well done

PR #1641, opened 2023-03-08
first pratice react

FallenMAD:


src/components/Header/Header.jsx
```diff
@@ -1,6 +1,16 @@
...
 
-// Create a Header function returning the HTML of header block
+function Header() {
+  return (
+    <>
+      <header className="header">
+        <h1 className="header__title">Site Name</h1>
+        <Navigation />
+      </header>
+    </>
```
polosanya, line 12: Do you really need a React fragment here? You only have one container, so it's redundant here 
Fix all cases

polosanya (changes requested):
Let's remove redundant React fragments. Add it only if you have two or more sibling elements

polosanya (approved):
Well done!

PR #1642, opened 2023-03-08
Solution

RuslanLiapin:
[DEMO LINK](https://ruslanliapin.github.io/react_decompose/)

### DIFF
--- a/src/App.jsx
+++ b/src/App.jsx
@@ -1,80 +1,15 @@
 import React from 'react';
 import './App.css';
-// Move each BEM block to a separate component (file) and import them here
+import Article from './article';
+import header from './header';
+import welcome from './welcome';
 
 function App() {
   return (
     <main className="app">
-      <section className="welcome">
-        <span className="welcome__text">Sticky Header!</span>
-      </section>
-      <header className="header">
-        <h1 className="header__title">Site Name</h1>
-        <nav className="navigation">
-          <a className="navigation__link" href="#about">About</a>
-          <a className="navigation__link" href="#services">Services</a>
-          <a className="navigation__link" href="#contact">Contact</a>
-        </nav>
-      </header>
-      <article className="article">
-        <h1 className="article__title">Headline</h1>
-        <p className="article__paragraph">
-          In elementum lorem eget est euismod ornare. Phasellus sit amet
-          pellentesque mauris. Aliquam quis malesuada ex. Nullam eu aliquam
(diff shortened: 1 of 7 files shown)


src/App.jsx
```diff
@@ -1,80 +1,15 @@
 import React from 'react';
 import './App.css';
-// Move each BEM block to a separate component (file) and import them here
+import Article from './Article';
```
etojeDenys, line 3:
```suggestion
import Article from './components/Article';
```

src/Navigation.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+const nav = (
+  <nav className="navigation">
+    <a className="navigation__link" href="#about">About</a>
+    <a className="navigation__link" href="#services">Services</a>
+    <a className="navigation__link" href="#contact">Contact</a>
+  </nav>
+);
```
etojeDenys, line 9: you need to rewrite it as a component, not variable

src/header.jsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import nav from './Navigation';
+
+const header = (
+  <header className="header">
+    {nav}
+  </header>
+);
```
etojeDenys, line 8: same here

src/welcome.jsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+const welcome = (
+  <section className="welcome">
+    <span className="welcome__text">Sticky Header!</span>
+  </section>
```
etojeDenys, line 6: here also

etojeDenys (changes requested):
and move all components into components folder

src/components/Header/Header.jsx
```diff
@@ -4,3 +4,14 @@
 // Create a Header function returning the HTML of header block
 
 // Add a default export statement for Header component to use it in the other files
+import React from 'react';
+import './Header.css';
+import nav from '../Navigation/Navigation';
+
+const Header = () => (
+  <header className="header">
+    {nav}
```
vlad-demchuk, line 13: it's not working now

```suggestion
    <Nav />
```

vlad-demchuk: you should use your component in this way

vlad-demchuk (changes requested):

sTorba24 (approved):
GJ 🚀

PR #1673, opened 2023-03-10
solution

Nordexz:
[DEMO LINK](https://Nordexz.github.io/react_decompose/)

### DIFF
--- a/src/components/Article/Article.jsx
+++ b/src/components/Article/Article.jsx
@@ -1,6 +1,66 @@
-// import the React library here
-// import a css file containig article styles
+import React from 'react';
+import './Article.css';
 
-// Create an Article function returning the HTML of article block
+const Article = () => (
+  <article className="article">
+    <h1 className="article__title">Headline</h1>
+    <p className="article__paragraph">
+      In elementum lorem eget est euismod ornare. Phasellus sit amet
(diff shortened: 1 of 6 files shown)


src/components/Article/Article.jsx
```diff
@@ -1,6 +1,66 @@
-// import the React library here
-// import a css file containig article styles
+import React from 'react';
+import './Article.css';
```
Andr1yK, line 2: Where is this file?
image


[Checklist](https://github.com/mate-academy/react_decompose/blob/master/checklist.md) point 3

Andr1yK (changes requested):

src/components/Navigation/Navigation.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import './Navigation.css';
+
+const Nav = () => (
```
Andr1yK, line 4:
```suggestion
// Nav.jsx
import './Nav.css';

const Nav = () => (
```

or 

```suggestion
import './Navigation.css';

const Navigation = () => (
```

Andr1yK: better to use same names for component and file

Andr1yK (approved):

PR #1675, opened 2023-03-10
react_decompose done

Stepan-TS:


src/components/Article/Article.css
```diff
@@ -1 +1,37 @@
...
+.article__paragraph:not(:first-child) {
+  margin-top: 0.7em;
+}
+
+@media (min-width: 500px) {
+  .header {
+    text-align: left;
+    display: flex;
+    justify-content: space-between;
+  }
```
Andr1yK, line 25: why do we need .header in the file for .article?

Andr1yK (approved):

PR #1677, opened 2023-03-10
Solution

Ihor-Zakharov:


src/components/Header/Header.jsx
```diff
@@ -4,3 +4,18 @@
...
+import './Header.css';
+
+const Header = () => (
+  <header className="header">
+    <h1 className="header__title">Site Name</h1>
+    <nav className="navigation">
+      <a className="navigation__link" href="#about">About</a>
+      <a className="navigation__link" href="#services">Services</a>
+      <a className="navigation__link" href="#contact">Contact</a>
+    </nav>
```
SanyaBratashchuk, line 17: Create navigation component

SanyaBratashchuk (changes requested):

Volodymyr-Melnyk-Fron10der (approved):
Good👍

PR #1701, opened 2023-03-24
React decompose

Oleksandr-Maslo:
https://oleksandr-maslo.github.io/react_decompose/


src/components/Header/Header.jsx
```diff
@@ -1,6 +1,14 @@
-// import the React library here
-// import a css file containig article styles
+import React from 'react';
+import './Header.css';
+import Navigation from '../Navigation/Navigation';
```
AlisaVlasova, line 3: add Navigation component to git as well

AlisaVlasova (changes requested):

AlisaVlasova (approved):

PR #1704, opened 2023-03-26
Solution

OlyaNikora:
[DEMO LINK](https://OlyaNikora.github.io/react_decompose/)


src/components/Header/Header.jsx
```diff
@@ -1,6 +1,13 @@
-// import the React library here
-// import a css file containig article styles
+import React from 'react';
+import Navigation from '../Navigation/Navigation';
```
VitaliyBondarenko1982, line 2: looks like you didn't push Navigation folder.
Screenshot from 2023-03-26 18-17-15
Screenshot from 2023-03-26 18-16-22

VitaliyBondarenko1982 (changes requested):

src/components/Header/Header.jsx
OlyaNikora: Hi, I've checked.  Navigation was pushed. Readme file was not pushed...
I 've added Readme

image

witflash (approved):
Good job! 👍🏻

PR #1707, opened 2023-03-27
Solution

IrynaK-a:
[DEMO LINK](https://IrynaK-a.github.io/react_decompose/)


package.json
```diff
@@ -1,12 +1,14 @@
 {
   "name": "react_decompose",
-  "homepage": ".",
+  "homepage": "http://.",
```
polosanya, line 3:
```suggestion
  "homepage": ".",
```
You don't need to change it. Your Demo link is not working now. This can be caused by changes in package.json

polosanya (changes requested):
Almost done!
The demo link is not working. Let's try to fix it.
Feel free to ask for help in fe_chat if you have any difficulties 🙂

PR #1708, opened 2023-03-27
Solution

yana-kotsulym:
https://yana-kotsulym.github.io/react_decompose/


src/components/Article/Article.jsx
```diff
@@ -1,6 +1,66 @@
...
-// import a css file containig article styles
+import React from 'react';
+import './Article.css';
 
-// Create an Article function returning the HTML of article block
+const Article = () => (
+  <article className="article">
+    <h1 className="article__title">Headline</h1>
+    <p className="article__paragraph">
+      In elementum lorem eget est euismod ornare. Phasellus sit amet
```
OleziO, line 8: Move text to the variable or object field.

OleziO: Check all other cases (better is to use object for all texts)

OleziO (changes requested):

IvanFesenko (approved):

PR #1709, opened 2023-03-28
solution

Spec0s:
https://spec0s.github.io/react_decompose/


src/components/Header/Header.jsx
```diff
@@ -1,6 +1,17 @@
...
-// Create a Header function returning the HTML of header block
+function Header() {
+  return (
+    <header className="header">
+      <h1 className="header__title">Site Name</h1>
+      <nav className="navigation">
+        <a className="navigation__link" href="#about">About</a>
+        <a className="navigation__link" href="#services">Services</a>
+        <a className="navigation__link" href="#contact">Contact</a>
+      </nav>
```
anstsot, line 12: according to the task you may create a folder for the `Navigation` component and move navigation there

Spec0s:
why does check fail ?

Spec0s:
but i dont get complete status ?

witflash (changes requested):
I think you forgot to add your new files (`Navigation.jsx` and `Navigation.css`).
Check it, please. Do the command `git add` for these files, then commit them and push to the repo.

polosanya (approved):
Good job!

PR #1718, opened 2023-04-01
Solution react_decompose

NazariiAlieksieiev:
[DEMO LINK](https://NazariiAlieksieiev.github.io/react_decompose/)


src/components/Article/Article.css
```diff
@@ -1 +1,36 @@
...
+  margin: 0;
+  color: #333;
+}
+
+.article__paragraph:not(:first-child) {
+  margin-top: 0.7em;
+}
+
+@media (min-width: 500px) {
+  .header {
```
witflash, line 20: This selector should be in header file

witflash: As I see you already have this styles there. So just remove this styles here

witflash (approved):
Good work 👍🏻

PR #1778, opened 2023-04-12
task done

andriy-litkovets:
[DEMO LINK](https://andriy-litkovets.github.io/react_decompose/)


src/components/Navigation/Navigation.jsx
```diff
@@ -0,0 +1,17 @@
+// import the React library here
```
agniya-i, line 1: not necessary to leave  comments)

agniya-i (approved):

PR #1812, opened 2023-04-28
Solution

SuaBy514:
[DEMO LINK](https://suaby514.github.io/react_decompose/)


src/components/Header/Header.jsx
```diff
@@ -1,6 +1,21 @@
...
 // Create a Header function returning the HTML of header block
+function Header() {
+  return (
+    <header className="header">
+      <h1 className="header__title">Site Name</h1>
+      <nav className="navigation">
+        <a className="navigation__link" href="#about">About</a>
+        <a className="navigation__link" href="#services">Services</a>
+        <a className="navigation__link" href="#contact">Contact</a>
+      </nav>
```
d-wasilewski, line 15: Move this to Another folder too

d-wasilewski (changes requested):

src/components/Header/Header.jsx
SuaBy514: That is out of the scope of the assignment, I will still do it though

kasiakubisiak (approved):

PR #1821, opened 2023-05-03
First React page

NadieinOleh:
https://NadieinOleh.github.io/react_decompose/


src/components/Article/Article.jsx
```diff
@@ -1,6 +1,70 @@
-// import the React library here
-// import a css file containig article styles
+import React from 'react';
 
-// Create an Article function returning the HTML of article block
+import './Article.css';
+import '../../styles/MediaQuery.css';
```
IvanFesenko, line 4: Push all required files

src/components/Header/Header.jsx
```diff
@@ -1,6 +1,14 @@
-// import the React library here
-// import a css file containig article styles
+import React from 'react';
+import Navigation from './Navigation/Navigation';
```
IvanFesenko, line 2: Push all required files

IvanFesenko (changes requested):
Not all used files was added

qoa11a (approved):

PR #1829, opened 2023-05-10
solution

JulyaPetrovskaya:
[DEMO LINK](https://JulyaPetrovskaya.github.io/react_decompose/)


src/components/Welcome/Welcome.css
```diff
@@ -1 +1,14 @@
-/* Put welcome styles here */
```
etojeDenys, line 1: Replace '+' with space at the beginning of the line.

etojeDenys (changes requested):
your demo link does not work

src/components/Article/Article.jsx
```diff
@@ -1,6 +1,68 @@
-// import the React library here
-// import a css file containig article styles
+import React from 'react';
+import './Article.css';
 
-// Create an Article function returning the HTML of article block
+function Article() {
+  return (
```
VitaliyBondarenko1982, line 5:
```suggestion
const Article = () => (
```
here and in other places can be arrow function.

VitaliyBondarenko1982 (approved):

PR #1840, opened 2023-05-20
solutuion

generatedman:


src/components/Article/Article.css
```diff
@@ -1 +1,36 @@
...
+.article__paragraph:not(:first-child) {
+  margin-top: 0.7em;
+}
+
+@media (min-width: 500px) {
+  .header {
+    text-align: left;
+    display: flex;
+    justify-content: space-between;
+  }
```
33Orange, line 24: Why you have `.header` styles inside `Article.css` ?

src/components/Header/Header.jsx
```diff
@@ -1,6 +1,16 @@
-// import the React library here
-// import a css file containig article styles
+import React from 'react';
+import './Header.css';
+import '../Article/Article.css';
```
33Orange, line 3: Why are you importing the article styles in the header?

33Orange (changes requested):
Almost done!

A few moments to fix

src/components/Welcome/Welcome.jsx
```diff
@@ -1,6 +1,16 @@
-// import the React library here
+import React from 'react';
+import './Welcome.css';
+
 // import a css file containig article styles
 
 // Create a Welcome function returning the HTML of welcome block
```
33Orange, line 6: Before pushing, you need to clean the code from unnecessary comments

33Orange (approved):
Well done

PR #1844, opened 2023-05-23
add task solution

maksym-nemera:
[DEMO LINK](https://maksym-nemera.github.io/react_decompose/)


src/components/Header/Header.jsx
```diff
@@ -1,6 +1,14 @@
...
 
-// Create a Header function returning the HTML of header block
+function Header() {
+  return (
+    <header className="header">
+      <h1 className="header__title">Site Name</h1>
+      <Navigation />
+    </header>
+  );
+}
```
ruslanpashkov, line 12: we can also use arrow functions for our components
```suggestion
const Header = () => (
  <header className="header">
    <h1 className="header__title">Site Name</h1>
    <Navigation />
  </header>
);
```

silchyn (changes requested):
Fix comment above and remove redundant comments

33Orange (approved):
Well done !

PR #1852, opened 2023-05-28
Solution

G37ua:
[DEMO LINK](https://g37ua.github.io/react_decompose/)


G37ua:
image
I don’t understand why my tests pass without problems, but here it gives errors

src/components/Article/Article.jsx
```diff
@@ -1,6 +1,70 @@
-// import the React library here
-// import a css file containig article styles
+import React from 'react';
+import './Article.css';
 
-// Create an Article function returning the HTML of article block
+function Article() {
+  return (
+    <>
```
TarasHoliuk, line 6: if you have one parent tag (`article` in this case) - react fragment is redundant
```suggestion
```

src/components/Header/Header.jsx
```diff
@@ -1,6 +1,16 @@
-// import the React library here
-// import a css file containig article styles
+import React from 'react';
+import Navigation from '../Navigation/Navigation';
```
TarasHoliuk, line 2: add this file to GitHub
Hint: use `git add .` or `git add src/.` to add created files before commit instead of `git commit -am`

src/components/Header/Header.jsx
```diff
@@ -1,6 +1,16 @@
...
+import './Header.css';
 
-// Create a Header function returning the HTML of header block
+function Header() {
+  return (
+    <>
```
TarasHoliuk, line 7: redundant react fragment

TarasHoliuk (changes requested):

TarasHoliuk:
> image I don’t understand why my tests pass without problems, but here it gives errors

It's GitHub issue not caused because of your code - nevermind (sometime it may happen)

OleziO (approved):

PR #1854, opened 2023-05-29
separete App component between Welcome, Header, Navigition, Article c…

AlexKherson:
…omponents
[DEMO LINK](https://alexKherson.github.io/react_decompose/)


src/components/Article/Article.jsx
```diff
@@ -1,6 +1,68 @@
-// import the React library here
-// import a css file containig article styles
+import React from 'react';
+import './Article.css';
 
-// Create an Article function returning the HTML of article block
+function Article() {
+  return (
```
naziamoff, line 5: you can use arrow syntax with a right-away return statement
```suggestion
const Article = ()  => (
```

src/components/Header/Header.jsx
```diff
@@ -1,6 +1,14 @@
-// import the React library here
-// import a css file containig article styles
+import React from 'react';
+import './Header.css';
+import Navigation from '../Navigation/Navigation';
 
-// Create a Header function returning the HTML of header block
+function Header() {
```
naziamoff, line 5: same here

src/components/Navigation/Navigation.jsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import './Navigation.css';
+
+function Navigation() {
```
naziamoff, line 4: and here

naziamoff (changes requested):
let's use arrow syntax

naziamoff (approved):
🔥

PR #1876, opened 2023-06-02
made task react decompose

vladyslav-lapshyn:
[DEMO LINK](https://vladyslav-lapshyn.github.io/react_decompose/)

I commented out some of the tests in the `header` block because they don't work because I added a separate block for navigation. I added new separate tests for navigation


src/App.jsx
```diff
@@ -1,80 +1,19 @@
...
-          tincidunt quis sodales vitae, ornare a turpis. Proin sit amet finibus
-          enim. Fusce tempus a neque vitae tempor. Aenean rutrum, libero
-          iaculis interdum vulputate, dui eros vehicula nisi, at interdum enim
-          lacus eu diam.
-        </p>
-      </article>
+      <Welcome />
+      <Header>
+        <Navigation />
+      </Header>
```
33Orange, line 15: Move navigation to header component.

src/components/Header/Header.jsx
```diff
@@ -1,6 +1,11 @@
...
-// import a css file containig article styles
+import React from 'react';
+import './Header.css';
 
-// Create a Header function returning the HTML of header block
+const Header = ({ children }) => (
+  <header className="header">
+    <h1 className="header__title">Site Name</h1>
+    {children}
+  </header>
```
33Orange, line 8:
```suggestion
  <header className="header">
    <h1 className="header__title">Site Name</h1>
    <Navigation />
  </header>
```

33Orange (changes requested):
Almost done.

Move navigation component to header and return the tests to their original place, because this is will fix the problems that were

PR #1905, opened 2023-06-06
done

yaroslav5481:
https://yaroslav5481.github.io/react_decompose/


src/components/Article/Article.jsx
```diff
@@ -4,3 +4,70 @@
 // Create an Article function returning the HTML of article block
 
 // Add a default export statement for Article component to use it in the other files
+import React from 'react';
+import './Article.css';
+
+const Article = () => (
+  <>
```
TarasHoliuk, line 11: `React.fragment` is needed if you want to return more than 1 parent tag
in this case you have **one** parent tag - `article`, so `React.fragment` is redundant

fix it everywhere

TarasHoliuk (changes requested):

TarasHoliuk (approved):

PR #1911, opened 2023-06-09
Solution

Ir-ra:
[DEMO LINK](https://Ir-ra.github.io/react_decompose/)


src/components/Article/Article.css
```diff
@@ -1 +1,36 @@
...
+  margin: 0;
+  color: #333;
+}
+
+.article__paragraph:not(:first-child) {
+  margin-top: 0.7em;
+}
+
+@media (min-width: 500px) {
+  .header {
```
witflash, line 20: I think it should be inside `Header.css`

witflash (changes requested):
Good job!
Just check my comment and make a fix

src/components/Header/Header.jsx
```diff
@@ -1,6 +1,15 @@
-// import the React library here
-// import a css file containig article styles
+import React from 'react';
+import './Header.css';
 
-// Create a Header function returning the HTML of header block
+const Header = () => (
+  <header className="header">
+    <h1 className="header__title">Site Name</h1>
+    <nav className="navigation">
```
agniya-i, line 7: Create component for navigation as well

agniya-i (approved):

PR #1922, opened 2023-06-16
add task solution

sasdaiv:
[DEMO LINK](https://sasdaiv.github.io/react_decompose/)


src/components/Header/Header.css
```diff
@@ -1 +1,27 @@
-/* Put header styles here */
+.header {
+  width: 100%;
+  padding: 1em;
+  font-size: 140%;
+  position: sticky;
```
etojeDenys, line 5: Consider adding a background color to the header for better visibility when using position: sticky.

src/components/Header/Header.css
```diff
@@ -1 +1,27 @@
...
+  top: 0;
+  left: 0;
+  right: 0;
+  transition: opacity 0.2s ease-in-out;
```
etojeDenys, line 9: It seems like the transition property is applied only to opacity, but there's no opacity property in this file. Make sure to add the necessary properties for the transition to work.

src/components/Welcome/Welcome.jsx
```diff
@@ -1,5 +1,10 @@
-// import a css file containig article styles
+import React from 'react';
+import './Welcome.css';
 
-// Create a Welcome function returning the HTML of welcome block
```
etojeDenys, line 3: Remove the comment '// Create a Welcome function returning the HTML of welcome block' as the code is self-explanatory.

src/components/Welcome/Welcome.jsx
```diff
@@ -1,5 +1,10 @@
-// import a css file containig article styles
```
etojeDenys, line 1: Remove the comment '// import a css file containig article styles' as it's not needed.

src/components/Welcome/Welcome.jsx
```diff
@@ -1,5 +1,10 @@
...
+import './Welcome.css';
 
-// Create a Welcome function returning the HTML of welcome block
+const Welcome = () => (
+  <section className="welcome">
+    <span className="welcome__text">Sticky Header!</span>
+  </section>
+);
 
-// Add a default export statement for Welcome component to use it in the other files
```
etojeDenys, line 5: Remove the comment '// Add a default export statement for Welcome component to use it in the other files' as the code is self-explanatory.

src/components/Header/Header.jsx
```diff
@@ -1,5 +1,12 @@
-// import a css file containig article styles
+import React from 'react';
+import './Header.css';
+import Navigation from '../Navigation/Navigation';
 
-// Create a Header function returning the HTML of header block
+const Header = () => (
+  <header className="header">
+    <h1 className="header__title">Site Name</h1>
+    <Navigation />
```
etojeDenys, line 8: you forgot to push Navigation to github

etojeDenys (changes requested):

etojeDenys (approved):
awesome

PR #1925, opened 2023-06-17
code

liza-strykharchuk:


Volodymyr-Melnyk-Fron10der (changes requested):
Fix demo link plz
image

Commit Navigation component plz as well

src/components/Article/Article.css
```diff
@@ -1 +1,36 @@
-/* Put article styles here */
+.article {
+  margin: 5em auto 0;
+  padding: 1em;
+  font-size: 140%;
+  max-width: 800px;
+  background: white;
+  box-shadow: rgba(0, 0, 0, 0.05) 0 3px 15px;
```
OleziO, line 7: Use HEX instead of RGBA

src/components/Article/Article.css
```diff
@@ -1 +1,36 @@
-/* Put article styles here */
+.article {
+  margin: 5em auto 0;
+  padding: 1em;
+  font-size: 140%;
+  max-width: 800px;
+  background: white;
```
OleziO, line 6: Don't use color names, use HEX codes.

OleziO (changes requested):

src/components/Article/Article.css
liza-strykharchuk: Ok, but my task was to split the App.js into components based on CSS blocks, this code was given to me in the start file...

vlad-demchuk (approved):

PR #1926, opened 2023-06-18
solution

RodionSav:
https://RodionSav.github.io/react_decompose/


src/components/Navigation/Navigation.jsx
```diff
@@ -0,0 +1,12 @@
+
+import './Navigation.css';
+
+const Nav = () => (
```
vlad-demchuk, line 4:
```suggestion
const Navigation = () => (
```

vlad-demchuk (changes requested):
demo link isn't working

PR #1942, opened 2023-06-24
solution

Absolution26:


polosanya (changes requested):
The code looks great, but please don't forget to add the demo link to the PR description, so we can check your page properly 🙂 
image

src/components/Header/Header.jsx
```diff
@@ -1,5 +1,17 @@
 // import a css file containig article styles
-
+import './Header.css';
 // Create a Header function returning the HTML of header block
 
 // Add a default export statement for Header component to use it in the other files
+const Header = () => (
+  <header className="header">
+    <h1 className="header__title">Site Name</h1>
+    <nav className="navigation">
```
witflash, line 9: It's better to move `navigation` to a separate component if you really have a different block.
Or you can just use class names as elements of `header`: `header__navigation`, `header__link`, etc. 
It's not critical but will improve your code.

witflash (approved):
Good job 👍

PR #1989, opened 2023-06-28
react-decompose solution

NataliGru:
- [DEMO LINK](https://NataliGru.github.io/react_decompose/)


parmenid89:
GJ, approve. But maybe it is better to use SCSS but not CSS.

src/components/Article/Article.jsx
```diff
@@ -1,5 +1,68 @@
-// import a css file containig article styles
+import React from 'react';
+import './Article.css';
 
-// Create an Article function returning the HTML of article block
+export function Article() {
```
ManoylovAC, line 4: you can use arrow function for components

ManoylovAC (approved):
Good work
Next time please use the arrow function for components

src/components/Article/Article.jsx
NataliGru: Okay, thank you!)

PR #2022, opened 2023-07-06
solution

Pa1eOrc:


src/components/Welcome/Welcome.css
```diff
@@ -1 +1,14 @@
-/* Put welcome styles here */
+.welcome {
+  height: 100vh;
+  background: coral;
```
OleziO, line 3: Use HEX instead of colors names.

OleziO (changes requested):

src/components/Welcome/Welcome.css
```diff
@@ -1 +1,14 @@
-/* Put welcome styles here */
+.welcome {
+  height: 100vh;
+  background: #ff7f50;
+}
+
+.welcome__text {
+  margin: 0;
+  font-size: 400%;
```
VitaliyBodnar, line 8: it's a bad practice to use % for font size, better to use px or rem for them

demetrxx (approved):
Well done 👍 Go ahead!

PR #2023, opened 2023-07-06
solution

artem-ivashchenko:
[DEMO LINK](https://artem-ivashchenko.github.io/react_decompose/)


src/components/Navigation/Navigation.jsx
```diff
@@ -0,0 +1,11 @@
+import './Navigation.css';
+
+export function Navigation() {
```
lerastarynets, line 3: The component is not exported as default. According to the task description, all components should use `export default`.

lerastarynets (changes requested):
Great job overall! Just one tiny thing needs to be fixed before approval)

lerastarynets (approved):
You're doing amazing, keep it up!

PR #2057, opened 2023-07-18
add task solution

NadiiaR:
[DEMO LINK](https://NadiiaR.github.io/react_decompose/)


src/components/Header/Header.jsx
```diff
@@ -1,5 +1,14 @@
-// import a css file containig article styles
+import './Header.css';
 
-// Create a Header function returning the HTML of header block
+export const Header = () => (
+  <header className="header">
+    <h1 className="header__title">Site Name</h1>
+    <nav className="navigation">
```
mykhalenych, line 6: create Navigation.jsx with this code

mykhalenych (changes requested):

src/components/Header/Header.jsx
NadiiaR: did not see this point in the task...

Viktor-Kost (approved):

PR #2059, opened 2023-07-20
solution

Vepr984:


IvanFesenko (changes requested):
Where is DEMO LINK?

Vepr984:
https://vepr984.github.io/react_decompose/

etojeDenys (changes requested):
1. You may also create a folder for the Navigation component and move navigation there.
2. remove all comments

src/components/Header/Header.jsx
```diff
@@ -1,5 +1,9 @@
-// import a css file containig article styles
+import './Header.css';
 
-// Create a Header function returning the HTML of header block
+const Header = () => (
+  <header className="header">
+    <h1 className="header__title">Site Name</h1>
```
etojeDenys, line 5: you need to put navigation here in header

etojeDenys (changes requested):

PR #2067, opened 2023-07-23
commit

anastasia-radchencko:
https://anastasia-radchencko.github.io/react_decompose/


src/components/Article/Article.css
```diff
@@ -1 +1,36 @@
...
+.article__paragraph:not(:first-child) {
+  margin-top: 0.7em;
+}
+  
+@media (min-width: 500px) {
+  .header {
+    text-align: left;
+    display: flex;
+    justify-content: space-between;
+  }
```
ivbutsykin, line 24: Probably it should be in `Header.css`

ivbutsykin (approved):
Well done!

Try to pass tests for yourself =)
For this you should make `default export` instead `export`: https://github.com/mate-academy/react_decompose/blob/master/README.md?plain=1#L6

PR #2082, opened 2023-07-28
done

Alessandranitti15:
[DEMO LINK](https://Alessandranitti15.github.io/react_decompose/)


src/components/Header/Header.css
```diff
@@ -1 +1,42 @@
...
+
+.header__title {
+  font-weight: 600;
+  display: inline;
+  margin: 0;
+  padding: 0;
+  font-size: inherit;
+}
+
+.navigation__link {
```
lerastarynets, line 21: The styles for `.navigation__link` should not be in the `Header.css` file. They should be moved to a new CSS file in the same directory as the `Navigation` component.

src/components/Header/Header.jsx
```diff
@@ -1,5 +1,14 @@
-// import a css file containig article styles
+import './Header.css';
 
-// Create a Header function returning the HTML of header block
+const Header = () => (
+  <header className="header">
+    <h1 className="header__title">Site Name</h1>
+    <nav className="navigation">
```
lerastarynets, line 6: The `Navigation` component is not separated into its own component and folder. It should be moved to a separate component as per the task description.

lerastarynets (changes requested):
I see you did a hard work) Let's improve it a little bit

lerastarynets (approved):
Great job)

PR #2084, opened 2023-07-29
init commit

domalewski:
[DEMO](https://domalewski.github.io/react_decompose/)


src/components/Welcome/Welcome.css
HaniaNassalska, line 1: You should remove the empty space 🙂

HaniaNassalska (changes requested):
You didn't copy all the styles to the right components. You should also remove them from the App.css file. 
You didn't create a Navigation component.

choeqq (approved):
Looks good to me ✅

PR #2136, opened 2023-08-16
Solution

szymonpachucki:
[DEMO LINK](https://szymonpachucki.github.io/react_decompose/)

### DIFF
--- a/src/components/Article/Article.jsx
+++ b/src/components/Article/Article.jsx
@@ -3,3 +3,71 @@
 // Create an Article function returning the HTML of article block
 
 // Add a default export statement for Article component to use it in the other files
+import './Article.css';
+import React from 'react';
+
+function Article() {
+  return (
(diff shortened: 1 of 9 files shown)


src/components/Article/Article.jsx
```diff
@@ -3,3 +3,71 @@
 // Create an Article function returning the HTML of article block
 
 // Add a default export statement for Article component to use it in the other files
+import './Article.css';
+import React from 'react';
```
Sancz0pansa, line 7: I think you don't need that here.

Sancz0pansa (changes requested):
I don't see App.css file here.

choeqq (changes requested):
I think you should leave `App.css` with default styles - until welcome styles

choeqq (approved):
Looks good to me ☑️

PR #2152, opened 2023-08-20
feat: add task solution

FrontKid:


Valentyn-Bilyk (approved):
I think it's perfect!

alexander-ignatow (approved):
Looks good! 👍 

One thing - on the resulting page you have multiple `h1` tags, which is not ok.
Make sure to mind this in the future!

PR #2179, opened 2023-08-21
add task solution

MateuszSlezak:
https://MateuszSlezak.github.io/react_decompose/


src/components/Article/Article.jsx
```diff
@@ -1,5 +1,69 @@
 // import a css file containig article styles
-
+import './Article.css';
 // Create an Article function returning the HTML of article block
```
PawelP2001, line 3: I would suggest to code accordingly to the comments if you decided to leave them, but other than that looks good to me!

PawelP2001 (approved):
Looks good

choeqq (changes requested):
Please remove all the comments and you're good to go 🚀

choeqq (approved):

PR #2181, opened 2023-08-21
add task solution

yuliiaant:
[DEMO LINK](https://yuliiaant.github.io/react_decompose/)


AngerDESTROYS (approved):
Good job!!!

src/App.jsx
```diff
@@ -1,79 +1,15 @@
 import React from 'react';
 import './App.css';
-// Move each BEM block to a separate component (file) and import them here
 
-const App = () => (
+import Header from './components/Header/Header';
+import Article from './components/Article/Article';
+import Welcome from './components/Welcome/Welcome';
+
+export const App = () => (
```
BudnikOleksii, line 8: Your App component is default exported, so don't combine named export in this case.

src/index.jsx
```diff
@@ -1,7 +1,7 @@
 import ReactDOM from 'react-dom';
-import App from './App';
+import Root from './App';
```
BudnikOleksii, line 2: Use the same name that your component has, if you rename it in this case it is a bit confusing.
The App is a common name for the root component so you don't need to rename it.

BudnikOleksii (changes requested):
Good job!
Left a few comments, check them.

tiserett (approved):

PR #2182, opened 2023-08-21
add task solution

vserhiichuk:
[DEMO LINK](https://vserhiichuk.github.io/react_decompose/)


serkrops (changes requested):
Hi! In this solution you don't need to use scss files.

vserhiichuk:
> Hi! In this solution you don't need to use scss files.

src/components/Article/Article.jsx
```diff
@@ -1,5 +1,67 @@
-// import a css file containig article styles
+import './Article.scss';
 
-// Create an Article function returning the HTML of article block
+function Article() {
+  return (
```
tiserett, line 4: by using function expression we can shorthand our code ;)

tiserett (changes requested):
make sure to replace all function declarations with expressions, this way we can shorthand our code ;)

Moroz-Dmytro (approved):

PR #2193, opened 2023-08-21
Develop

dpidlutska:
[DEMO LINK](https://dpidlutska.github.io/react_decompose/)


src/components/Article/Article.jsx
```diff
@@ -1,5 +1,65 @@
-// import a css file containig article styles
+import './Article.css';
 
-// Create an Article function returning the HTML of article block
+export const Article = () => (
```
tiserett, line 3: don't mix exports

src/components/Header/Header.jsx
```diff
@@ -1,5 +1,13 @@
-// import a css file containig article styles
+/* eslint-disable import/no-named-as-default */
 
-// Create a Header function returning the HTML of header block
+import './Header.css';
+import Navigation from '../Navigation/Navigation';
 
-// Add a default export statement for Header component to use it in the other files
+export const Header = () => (
```
tiserett, line 6: don't mix exports ;)

src/components/Navigation/Navigation.jsx
```diff
@@ -0,0 +1,11 @@
+import './Navigation.css';
+
+export const Navigation = () => (
```
tiserett, line 3: same here

tiserett (changes requested):
make sure not to mix your exports, in this case we use 'default' for tests ;)

tiserett (approved):

PR #2195, opened 2023-08-21
solution

Taras2705:


AntonHerasymov (approved):
LGTM 🔥

src/components/Header/Header.jsx
```diff
@@ -1,5 +1,11 @@
...
+import './Header.css';
+import Navigation from '../Navigation/Navigation';
 
-// Create a Header function returning the HTML of header block
+const Header = () => {
+  <header className="header">
+    <h1 className="header__title">Site Name</h1>
+    <Navigation />
+  </header>;
+};
```
Moroz-Dmytro, line 9: Looks like something is wrong here, that's why tests failed. What does your function return?

src/components/Welcome/Welcome.jsx
```diff
@@ -1,5 +1,10 @@
-// import a css file containig article styles
+import './Welcome.css';
 
-// Create a Welcome function returning the HTML of welcome block
+const Welcome = () => (
+  <section className="welcome">
+    <span className="welcome__text">Sticky Header!</span>
+  </section>
```
Moroz-Dmytro, line 7:
```suggestion
```

Moroz-Dmytro (changes requested):
Let's fix tests

src/components/Header/Header.jsx
Taras2705: I just noticed that I put the wrong brackets. thank you.

PR #2197, opened 2023-08-21
feat: add complite test

ArturShynkaryk:


src/components/Welcome/Welcome.jsx
```diff
@@ -3,3 +3,15 @@
 // Create a Welcome function returning the HTML of welcome block
 
 // Add a default export statement for Welcome component to use it in the other files
+
+import './Header.css';
```
etojeDenys, line 7: why do you need this import in Welcome component

etojeDenys (changes requested):
don't forget to add the demo link to PR description

PR #2202, opened 2023-08-22
add task solution

AntonHerasymov:
[DEMO LINK](https://AntonHerasymov.github.io/react_decompose/)


src/components/Article/Article.css
```diff
@@ -1 +1,32 @@
 /* Put article styles here */
```
Moroz-Dmytro, line 1: Don't leave comments in your code

src/components/Article/Article.jsx
```diff
@@ -1,5 +1,67 @@
-// import a css file containig article styles
+import './Article.css';
 
-// Create an Article function returning the HTML of article block
+function Article() {
```
Moroz-Dmytro, line 3: Better to use function expression, than you will be able to simplify this

Moroz-Dmytro: the same comment to all components

Moroz-Dmytro (changes requested):
GJ 👍 
But let's improve it

src/App.jsx
```diff
@@ -1,79 +1,14 @@
 import React from 'react';
 import './App.css';
-// Move each BEM block to a separate component (file) and import them here
+import Article from './components/Article/Article';
+import Header from './components/Header/Header';
+import Welcome from './components/Welcome/Welcome';
```
k1NAs, line 6: I think it would be better to import the components in the order you add them to the App component

<main className="app">
    <Welcome />
    <Header />
    <Article />
</main>

k1NAs (changes requested):
GJ :fire:
But from my point there is one little fix)

BudnikOleksii (approved):
Good job!

PR #2209, opened 2023-08-22
react_decompose

vatatan:
[DEMO LINK](https://vatatan.github.io/react_decompose/)


src/components/Header/Header.css
```diff
@@ -1 +1,32 @@
...
+}
+
+@media (min-width: 500px) {
+  .header {
+    text-align: left;
+    display: flex;
+    justify-content: space-between;
+  }
+
+  .article {
```
Yevheniia-Sid, line 28: You don't need CSS rule for .article in Header

Yevheniia-Sid (changes requested):
Good job!👍

But I think you should change your Header.css a bit

src/components/Header/Header.css
vatatan: thanks for the comment😊
corrected

tiserett (approved):

PR #2249, opened 2023-08-30
Solution

andrii-rykal:
[DEMO LINK](https://andrii-rykal.github.io/react_decompose/)


src/components/Header/Header.jsx
```diff
@@ -1,5 +1,13 @@
-// import a css file containig article styles
+import './Header.css';
+import Navigation from '../Navigation/Navigation';
```
IvanFesenko, line 3: Commit and Push all files
2023-08-30_13h29_03

IvanFesenko (changes requested):
Good job. But not all files was added

Viktor-Kost (approved):

PR #2970, opened 2024-05-15
Solution

lishvets:
[DEMO LINK](https://lishvets.github.io/react_decompose/) and add it to the PR description


etojeDenys (changes requested):
demo link does not work. please fix it and request the review again

PR #3522, opened 2025-02-25
add task solution

Olivio-Neves:
(https://<Olivio-Neves>.github.io/react_decompose/)


DanielCamposDevX (changes requested):
No demo link provided, remember to always provide a demo link

PR #3534, opened 2025-02-27
Develop

PiekharievaK:
[DEMO LINK](https://PiekharievaK.github.io/react_decompose/)


src/App.jsx
```diff
@@ -1,86 +1,15 @@
 import React from 'react';
 import './App.css';
-// Move each BEM block to a separate component (file) and import them here
+import Article from './components/Article/Article';
+import Header from './components/Header/Header';
+import Welcome from './components/Welcome/Welcome';
```
nataliia-v, line 5: it would be good to fix spaces between imports.

src/components/Header/Header.jsx
```diff
@@ -1,5 +1,13 @@
-// import a css file containig header styles
+import './Header.css';
+import Navigation from '../Navigation/Navigation';
```
nataliia-v, line 2: Firstly, we import components, and after components – local styles.

1. Imports from global packages (like 'react')
// add space
2. Import components
// add space
3. Import local styles

I can't find this information in the documentation, but it's a common rule in projects. We can ask the mentor about it.

nataliia-v (approved):

PR #3607, opened 2025-03-22
Develop

kevinlino1908:
[DEMO LINK](https://kevinlino1908.github.io/react_decompose/)

### DIFF
--- a/src/components/Header/Header.jsx
+++ b/src/components/Header/Header.jsx
@@ -1,5 +1,26 @@
 // import a css file containig header styles
 
+import './Header.css';
+
 // Create a Header function returning the HTML of header block
+const Header = () => {
+  return (
+    <header className="header">
+      <h1 className="header__title">Site Name</h1>
+      <nav className="navigation">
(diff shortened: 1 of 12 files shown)


kevinlino1908:
@mateacademy-ai-mentor todos os testes passaram

pedro-ruas (changes requested):
I'm not sure, but I believe the header should not take the whole screen height

src/components/Header/Header.jsx
```diff
@@ -1,5 +1,26 @@
...
+        <a className="navigation__link" href="#about">
+          About
+        </a>
+        <a className="navigation__link" href="#services">
+          Services
+        </a>
+        <a className="navigation__link" href="#contact">
+          Contact
+        </a>
+      </nav>
```
raulriato, line 20: You may also create a folder for the Navigation component and move navigation there.

raulriato (changes requested):
Your code looks great. Just need a little chnge.

PR #3717, opened 2025-05-13
added components

torry455:


yelyzavetaz (approved):
Great work on this — the implementation is clean and everything works as expected! ദ്ദി/ᐠ｡‸｡ᐟ\

One small suggestion from the task: it might be a good idea to extract the navigation into a separate Navigation component for better reusability and clarity. 

But overall, really solid job!

PR #3736, opened 2025-05-14
add task solution

Dikanb:
https://Dikanb.github.io/react_decompose/

Split the `App.js` into components based on CSS blocks. CSS code should be split too.
- You already have `Article`, `Header` and `Welcome` folders inside `./src/components/` with required files.
- You may also create a folder for the `Navigation` component and move navigation there.
- Tests expect you to use `export default` for all the components.


YarRudenko (changes requested):
You have already nice code, but your DEMO Link - has some problem and get Error 404.

PR #3759, opened 2025-05-24
React decompose

SasaOstojic:


vadim-ilchenko (approved):
Auto-approved: Kodree task fix

PR #3904, opened 2025-07-29
Solution

romkas228:
[DEMO LINK](https://romkas228.github.io/react_decompose/)


horidor (approved):
Good job! Code correctly decomposes react components into small modules that are easy to work with without modyfing the initial page design.

PR #3910, opened 2025-07-30
added task solution

horidor:
Link: [DEMO LINK](https://horidor.github.io/react_decompose/)


YuliaProkipchuk (approved):
Good job! In the future it would be better to remove comments.

PR #3915, opened 2025-07-30
solution

isobol-js:


src/components/Article/Article.jsx
```diff
@@ -1,5 +1,68 @@
-// import a css file containig article styles
+import React from 'react';
```
YevheniiKa, line 1: Starting with React 17 and newer, the compiler automatically adds the React import to the generated code, so it is not necessary to explicitly write import React from ‘react’.

PR #4010, opened 2025-09-15
Solution

IvanVaniuk:
[DEMO LINK](https://IvanVaniuk.github.io/react_decompose/)


src/components/Article/Article.css
```diff
@@ -1 +1,31 @@
 /* Put article styles here */
```
volodymyr-soltys97, line 1: Remove all comments
```suggestion
```

src/components/Article/Article.jsx
```diff
@@ -1,5 +1,68 @@
 // import a css file containig article styles
-
+import './Article.css'
 // Create an Article function returning the HTML of article block
-
+const Article = () => {
+  return (
+    <article className="article">
```
volodymyr-soltys97, line 6: You don't need to use return here
```suggestion
const Article = () => (
    <article className="article">
```

volodymyr-soltys97 (approved):
Good job 👍

PR #4375, opened 2026-04-21
Solution

yana-karpovych:
https://yana-karpovych.github.io/react_decompose/


DmitriyS2k (approved):
The task is solved, well done!

PR #4377, opened 2026-04-22
add task solution

DmitriyS2k:
[DEMO LINK](https://DmitriyS2k.github.io/react_decompose/)


src/components/Article/Article.jsx
```diff
@@ -1,5 +1,67 @@
-// import a css file containig article styles
+import React from 'react';
```
linska, line 1: import React is no longer needed here 🙂 With the new JSX transform, it’s handled automatically, so you can safely remove it to keep the code a bit cleaner.

linska (approved):
Looks good to me 🔥

DanilWeda (approved):

PR #4384, opened 2026-04-22
Develop

Evgeniy45:
DEMO LINK - https://Evgeniy45.github.io/react_decompose/


Viqsol:
Good job. It seems like all tasks were solved. You even remembered about link, and I forgot.

Viqsol (approved):

PR #4385, opened 2026-04-22
Decompose App into Welcome, Header, Article components

MaxShaggy:
https://MaxShaggy.github.io/react_decompose/


src/components/Header/Header.jsx
```diff
@@ -1,5 +1,23 @@
-// import a css file containig header styles
+import React from 'react';
+import './Header.css';
 
-// Create a Header function returning the HTML of header block
+function Header() {
+  return (
+    <header className="header">
+      <h1 className="header__title">Site Name</h1>
+      <nav className="navigation">
```
CyborgNinjaHat: The navigation should be moved into a separate component in accordance with the task requirements.

src/components/Header/Header.css
```diff
@@ -1 +1,42 @@
...
+
+.header__title {
+  font-weight: 600;
+  display: inline;
+  margin: 0;
+  padding: 0;
+  font-size: inherit;
+}
+
+.navigation__link {
```
CyborgNinjaHat: The styles for the navigation should be moved into a separate Navigation component.

CyborgNinjaHat (changes requested):
According to the requirements, you were supposed to move the navigation to a separate component.

PR #4390, opened 2026-04-23
git add solution

andriy-stetsula:


bstroshek (approved):
Everything looks good overall 👍  
One small suggestion: I would consider removing the separate Welcome.jsx file from src if it's not necessary.

PR #4396, opened 2026-04-24
add solution

Dima2702:


Palfinger2002 (approved):
Okey, good jod!!!

In Header.jsx, Article.jsx, and Welcome.jsx the components use function Component() {}, but Navigation inside Header.jsx uses an arrow function (const Navigation = () => {}). Both ways work, but it would be better to use the same style everywhere for consistency.

